PR #594, opened 2026-02-13
Unify orchestrator and CLI container launch via shared config builder

jwbron:
## Summary

Fixes orchestrator pipeline container launch failures by unifying the container configuration path with the CLI. The orchestrator was building container configs independently, causing multiple networking, mount, and environment bugs that prevented containers from functioning correctly.

### Commit 1: Fix pipeline container networking, logging, and contract init

- **Network mode**: Containers always launched on `egg-isolated` regardless of mode. Public-mode (issue) pipelines now use `egg-external` with correct gateway/orchestrator IPs
- **Missing CONTAINER_ID**: Orchestrator never set `CONTAINER_ID` env var, breaking gateway worktree path mapping (git wrapper sent "unknown" as container_id)
- **Contract init**: Failures were silently swallowed, leaving pipelines without contracts. Now fails fast with cleanup
- Subprocess stderr captured to log file; container log tail increased to 200 lines
- Entrypoint reads `EGG_PRIVATE_MODE` env var instead of inferring from `HTTPS_PROXY`

### Commit 2: Fix worktree prune destroying gateway worktrees

- `StateStore` called `git worktree prune` at init, which saw all gateway-managed worktrees as stale (orchestrator doesn't mount `/home/egg/.egg-worktrees`) and deleted their admin dirs
- Replaced with targeted `_remove_stale_admin_dir()` that only removes the state worktree's admin dir

### Commit 3: Unify orchestrator and CLI container launch via shared config builder

- Introduces `build_sandbox_config()` + `to_dockerpy_kwargs()` in `shared/egg_container/` so both CLI and orchestrator produce identical container configurations
- Fixes containers missing: `.git` shadow mounts, `--add-host` for gateway hostname, DNS lockdown for private mode, proxy env vars, and correct `GATEWAY_URL`
- Fixes `GATEWAY_URL` being overwritten by the now-removed `get_container_env()` in `gateway_client.py`
- Uses Docker API `mounts` list format (instead of legacy `volumes` dict) to support multiple repos without key collision
- `.git` shadow uses `/dev/null` bind mount (file-over-file) for worktrees, fixing the "not a directory" OCI runtime error when tmpfs was mounted over a worktree `.git` file
- Refactors `sandbox/egg_lib/runtime.py` to use shared `git_shadow_mounts()`

## Test plan

- [x] 42 new config builder tests pass (`tests/shared/egg_container/test_config_builder.py`)
- [x] 27 existing CLI command builder tests pass (no regressions)
- [x] 31 gateway client tests pass (removed obsolete `get_proxy_config`/`get_container_env` tests)
- [x] 210 non-docker-dependent orchestrator tests pass (updated spawner tests for new signature)
- [ ] Deploy and run `egg --local` — verify CLI behavior unchanged
- [ ] Deploy and run pipeline via orchestrator — verify `.git` shadowing, gateway hostname resolution, proxy/DNS for private mode

### DIFF
--- a/sandbox/entrypoint.py
+++ b/sandbox/entrypoint.py
@@ -1316,8 +1316,16 @@ def _tee_stderr_to_file(
         with open(log_path, "wb") as log_file:
             for saved_line in ring:
                 log_file.write(saved_line)
-    except Exception:
-        pass
+    except Exception as exc:
+        # Best-effort diagnostic — log so failures aren't completely silent.
+        try:
+            print(
+                f"[DEBUG] _tee_stderr_to_file failed: {exc}",
+                file=sys.stderr,
+            )
+            sys.stderr.flush()
+        except Exception:
+            pass
 
 
 def _run_with_stderr_capture(
--- a/shared/egg_container/__init__.py
+++ b/shared/egg_container/__init__.py
@@ -288,6 +288,14 @@ def to_dockerpy_kwargs(config: SandboxContainerConfig) -> dict:
         kwargs["dns"] = list(config.dns)
     if config.command:
         kwargs["command"] = list(config.command)
+    if config.container_ip:
+        kwargs["networking_config"] = {
+            "EndpointsConfig": {
+                config.network.network_name: {
+                    "IPAMConfig": {"IPv4Address": config.container_ip},
+                },
+            },
+        }
 
     return kwargs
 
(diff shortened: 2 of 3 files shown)


jwbron:
@james-in-a-box address the above feedback

jwbron:
@james-in-a-box address this feedback: https://github.com/jwbron/egg/pull/594#pullrequestreview-3794939196

PR #703, opened 2026-02-15
Fix container naming conflict when pipelines are restarted

jwbron:
## Summary
Fixes the Docker 409 Conflict error that occurs when a pipeline is canceled and restarted.

## Problem
When a pipeline was canceled and then restarted, old containers were not being cleaned up, causing Docker to throw a 409 Conflict error:
```
Failed to create container: 409 Client Error for http+docker://localhost/v1.52/containers/create?name=egg-sandbox-egg-issue-659-refiner: Conflict ("Conflict. The container name "/egg-sandbox-egg-issue-659-refiner" is already in use by container "c5f83245a79694872602b957a9a6bc07ed133c6bda5b082fc6db1d7eab1973a6". You have to remove (or rename) that container to be able to reuse that name.")
```

## Solution
Implemented three-layer cleanup to ensure containers are always properly removed:

### 1. Pre-emptive cleanup in `container_spawner.py`
Before creating a new container, check if a container with the same name already exists and remove it. This handles the case where a previous pipeline run left containers behind.

### 2. Cleanup on pipeline deletion
When a pipeline is deleted via the DELETE endpoint, all its containers are now properly removed before deleting the pipeline state.

### 3. Cleanup on pipeline cancellation
When a pipeline status is updated to CANCELLED or FAILED via the PATCH endpoint, all its containers are now cleaned up automatically.

## Test plan
- [ ] Cancel a running pipeline
- [ ] Restart the same pipeline
- [ ] Verify no 409 Conflict error occurs
- [ ] Verify old containers are properly removed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/orchestrator/routes/pipelines.py
+++ b/orchestrator/routes/pipelines.py
@@ -12,6 +12,7 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any
 
+from docker.errors import DockerException
 from flask import Blueprint, Response, jsonify, request, stream_with_context
 
 # Add shared directory to path for egg_logging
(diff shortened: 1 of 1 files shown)


jwbron:
@james-in-a-box address the above feedback

Issue #2140, opened 2026-04-27
request_changes on phase_gate strands pipeline due to stale pipeline-state worktree


jwbron:
## What happened

During an SDLC pipeline run on issue `issue-2137`, the operator chose `request_changes` on the refine-phase `phase_gate` decision. The orchestrator accepted the resolution but failed to spawn a new refine-phase cycle because of a stale worktree:

```
Git command failed: Preparing worktree (checking out 'egg/pipeline-state')
fatal: '/home/egg/.egg-state/pipeline-worktree-egg' already exists
```

Pipeline state went from `awaiting_human` → `failed` with `running_agents=[]`, `containers=[]`, `pending_decisions=[]`. No automatic recovery path; the operator had to cancel the pipeline manually (losing all refine-phase work).

## Reproduction

1. Submit any pipeline that reaches a refine-phase `phase_gate` decision.
2. On the gate, call `provide_input` with `{"action": "request_changes", "feedback": "..."}`.
3. Observe: status flips to `failed`; `get_pipeline_snapshot` shows `error: "Git command failed: ... '/home/egg/.egg-state/pipeline-worktree-egg' already exists"`.

## Expected behavior

`request_changes` should reset the phase and respawn the refiner team for a new cycle. The stale worktree from cycle 0 should be pruned (or reused) before cycle 1 attempts checkout.

## Suspected affected paths

- The failing checkout target is `/home/egg/.egg-state/pipeline-worktree-egg` — the orchestrator's pipeline-state worktree, distinct from container worktrees.
- `prune_stale_worktrees` does not clean this up because it's not a container worktree under `~/.egg-worktrees`.
- The phase-reset code path (likely the `request_changes` / `change_approach` branch in `orchestrator/routes/phases.py::advance_phase` or wherever the resolution is processed) is not reaping the prior cycle's pipeline-state worktree before re-checkout.
- `change_approach` likely has the same symptom — it shares the phase-reset code path with `request_changes`.

## Severity

Medium. Blocks the `request_changes` and likely `change_approach` HITL flows on phase gates — operators can resolve `approve` cleanly, but the negative paths strand the pipeline. Workaround: cancel and re-submit (loses all work).

## Notes

- Observed 2026-04-27 against pipeline `issue-2137` while testing `/sdlc` HITL.
- Orchestrator status: `healthy`. Gateway status: `healthy`. Code-path bug, not infra.
[...]

PR #2172, opened 2026-04-27
fix: state-store self-heal + surface wedges as 500 (#2167)

### LINKED ISSUE #2167: Orchestrator: stale prunable worktree on egg/pipeline-state branch makes every state-load 500 (surfaces as MCP 404)
## Symptom

While running `/sdlc` on issue #1924, ~50 minutes into the implement phase (BRC cycle 1, after `reviewer_contract` NACKed on a `ruff format` failure and `coder` had pushed the format-fix commit `33f2882a9`), every MCP call against pipeline `issue-1924` started returning `HTTP 404 NOT FOUND`:

- `mcp__egg__get_status(task_id="issue-1924")` → 404
- `mcp__egg__wait_for_status_change(task_id="issue-1924", ...)` → 404
- `mcp__egg__list_tasks()` returns a different unrelated pipeline (`KORE-1191-complete`) but **does not** list `issue-1924`.

Meanwhile:

- `mcp__egg__check_health` reports both `orchestrator` and `gateway` `healthy=true`, with `last_unhealthy_at: null` and no transitions during the wedge window.
- `ps -ef` still shows ~5 live agent containers for pipeline `issue-1924` — the `overseer`, `tester`, and the various `reviewer_*` and `documenter` agents whose system prompts hard-code `Pipeline ID: issue-1924`. They are wait-looping on `egg-orch message wait-loop` waiting for BRC events that will never come.
- The branch `origin/egg/issue-1924` is intact and has all 16 commits — analysis (`27ac5b150`), plan (`42fbb6433`), 6 phases of implementation, plus the ruff format fix (`33f2882a9`).
- Local pipeline state for `issue-1924` is fully gone from `.egg-state/pipelines/issue-1924/` (directory listing is empty), no contract files for the issue, no brc-history entries, no oversight session files.

So the pipeline state was deleted out from under the running agents while consensus was in progress. The agents will eventually time out, max-turns out, or be killed; until then they will fruitlessly retry against a vanished pipeline.

## Reproduction

I cannot give a tight repro because nothing in my session deliberately triggered cleanup. The sequence of operator actions on `issue-1924` before the wedge:

1. `submit_task(description=..., repo=..., issue_number=1924)` — Phase 1 + 2 ran clean.
2. Plan-phase gate `decision-22` resolved with `{\"action\":\"approve\"}` via `provide_input`.
3. The pipeline got stuck after that approve (logged separately as #2155). Operator called `advance_phase(target_phase=\"implement\")` — **without `force=true`** — to push it forward. That returned `{\"success\": true, \"data\": {\"current_phase\": \"implement\", \"previous_phase\": \"plan\"}}` and the implement-phase containers spawned within seconds.
4. Implement phase ran a clean BRC cycle. coder hit a git-policy wedge (logged separately as #2162); operator nudged via `send_message`; coder proposed cleanly against `aba4779b0`.
5. `reviewer_contract` NACKed on a `ruff format --check` failure with surgical line-by-line guidance.
6. coder pushed `33f2882a9 style(#1924): apply ruff format to jira_client + gateway` to address the NACK.
7. ~30 seconds later, every MCP call against `issue-1924` started returning 404.

A few candidate triggers worth chasing:

- **Did `advance_phase` from step 3 register the implement-phase pipeline run differently** in a way that stays visible in `list_tasks` but goes through a separate cleanup path? The fact that `list_tasks` does NOT list `issue-1924` despite live containers strongly suggests the pipeline registry entry is gone, not just the file-on-disk state.
- **Did the post-NACK push of `33f2882a9` trigger a branch-state reconciliation** that cascaded into a \"pipeline cleaned up\" code path?
- **Was the in-process orchestrator pipeline registry (e.g. `_pipelines: dict`) accidentally evicted** by a TTL / LRU / cleanup task while containers were still running?

## What I'd expect

- A live pipeline with running agent containers should never be removed from the registry.
- If the cleanup path runs anyway, it should at minimum stop the containers it is orphaning (so they don't burn cycles wait-looping into a void).
- `list_tasks` should not show an empty / unrelated set immediately after `advance_phase` reports success on the same pipeline.

## Diagnostic data

- Branch tip at the moment of the wedge: `33f2882a98f58aba390738e7fe92d7a3a61a2692` on `origin/egg/issue-1924`.
- Live container processes (`ps -ef`) at the time of the wedge include the `overseer` (PID 604335), `tester` (PID 604697), and additional `reviewer_*` containers — each with system prompts referencing `Pipeline ID: issue-1924`.
- `mcp__egg__check_health` snapshot at wedge: orchestrator + gateway both healthy, no transitions in the wedge window.
- `mcp__egg__list_tasks` at wedge: returned exactly one unrelated task (`KORE-1191-complete`).
- `mcp__egg__get_status(task_id=\"issue-1924\")` at wedge: `HTTP 404 NOT FOUND`.
- `.egg-state/pipelines/issue-1924/` does not exist at wedge time.
- `.egg-state/oversight/agent-timing.json` is present and 67 bytes (likely empty / fresh) — but no per-pipeline overrides for `issue-1924`.

## Workaround

Operator opens a draft PR from `origin/egg/issue-1924` (the branch is intact and has the bulk of the work). The orphaned containers will eventually exit on their own; the cost is wasted token spend until then.

## Asks

- A regression or property test: \"a pipeline whose containers are running cannot be removed from the registry\".
- If the cleanup is intentional under some condition (a stale-pipeline reaper, e.g.), make it explicit in the audit log so operators can attribute the disappearance.
- Surface the wedge: when `get_status` returns 404 for a pipeline that has live containers, the gateway / orchestrator side should at least log a warning. Right now this state is silent on the API surface.

Logged from `/sdlc` run on `2026-04-27`. Related to but distinct from #2155 (auto-advance after plan-phase gate failed) and #2162 (gateway git policy left coder with no recovery path after a detached-HEAD commit).

jwbron:
Closes #2167.

## Summary

Three fixes for the orchestrator-wide wedge in #2167 — the state branch was pinned by a stale prunable worktree, every state-load 500'd, and the failure surfaced to operators as 404 "Pipeline not found" with `check_health` still reporting `healthy=true`.

- **`orchestrator/state_store.py`** — self-heal `git worktree add` failures of the form `'egg/pipeline-state' is already used by worktree at '<X>'`. Parse `X` from stderr, verify it's gone from disk (never touch a live worktree), drop the single admin dir whose `gitdir` matches `X`, retry once. This catches the case the existing fix from #2140 doesn't: when the legacy admin dir at `pipeline-worktree` survives a deployment switch to `pipeline-worktree-{repo}`, the path-equality check in `_remove_stale_admin_dir` fails to match and cleanup is skipped.
- **`orchestrator/routes/pipelines.py`** — `_resolve_pipeline` no longer catches `StateStoreError` in its multi-repo loop. `GitOperationError` is a `StateStoreError` subclass; the broad except re-raised every git wedge as `PipelineNotFoundError` (→ 404). Now only `PipelineNotFoundError` is caught; infrastructure failures propagate. Combined with a new app-level error handler in `api.py`, callers see `500` with the actual error message and `error_type` field instead of "Internal server error".
- **`orchestrator/routes/health.py`** — `/api/v1/health` now probes the state-store live (`_ensure_worktree`). On failure, top-level `status` flips to `degraded` and the failing error message lands in `components.state_store`. HTTP stays `200` so k8s probes (`/live`, `/ready`) are unaffected. MCP `check_health` already reads the `status` field, so it now reflects the wedge.

## Out of scope

- `_collect_all_pipelines` still silently `continue`s on `StateStoreError` per repo (the side observation that `list_tasks` returned an unrelated Jira-driven pipeline during the wedge). Once the self-heal kicks in, list_tasks recovers; touching the broader silent-skip behavior is a separate decision.

## Test plan

- [x] `TestBranchHeldByPrunableWorktree` (4 cases) — recovery, safety guard against pruning live worktrees, unrelated-error propagation, path-independent admin-dir lookup.
- [x] `TestResolvePipelinePropagatesGitErrors` — `GitOperationError` is no longer translated to `PipelineNotFoundError`; legitimate not-found still surfaces correctly.
- [x] `TestStateStoreErrorHandlerSurfacesDetail` — end-to-end via Flask test client confirms 500 + actual error message + `error_type` field on the wire.
- [x] `TestHealthEndpointReflectsStateStoreFailure` — `/health` flips to `degraded` when probe fails; stays `healthy` when it succeeds; gracefully skipped on missing config / no request context.
- [x] Existing `/api/v1/health` test assertions in `test_health_check_integration.py` and `test_health_check_lifecycle_integration.py` patched to mock the probe (covered independently in the new file).
- [x] Targeted subset (170 tests across state_store, health, pipelines routes) passes; broader run prior to the test-ordering fix passed at 4753 (1 skipped).

### DIFF
--- a/orchestrator/api.py
+++ b/orchestrator/api.py
@@ -145,27 +145,30 @@ def log_request_end(response: Response) -> Response:
 @app.errorhandler(Exception)
 def handle_unhandled_exception(e: Exception) -> tuple[Response, int]:
     """Return JSON for all unhandled exceptions."""
+    from state_store import GitOperationError
     from werkzeug.exceptions import HTTPException
 
-    # Surface state-store infrastructure failures (e.g. ``git worktree``
+    # Surface state-store infrastructure failures (``git worktree``
     # contention) with the actual error so operators can diagnose without
     # grepping logs.  Routes used to swallow ``GitOperationError`` as
     # ``PipelineNotFoundError`` and return 404, masking a recoverable
     # wedge as "missing pipeline" (#2167).
-    try:
-        from state_store import StateStoreError
-
-        if isinstance(e, StateStoreError):
-            logger.error("State store error", error=str(e), exc_info=True)
-            return jsonify(
-                {
-                    "success": False,
-                    "message": f"State store unavailable: {e}",
-                    "error_type": type(e).__name__,
-                }
-            ), 500
-    except ImportError:  # pragma: no cover — defensive
-        pass
+    #
+    # Narrowed to ``GitOperationError`` deliberately: other ``StateStoreError``
+    # subclasses (``PipelineNotFoundError``, ``InvalidPipelineIdError``,
+    # ``StateValidationError``, ``VersionConflictError``) have correct
+    # status codes when handled at the route level; catching them here
+    # would surface a future "forgot to handle PipelineNotFoundError" as
+    # 500 instead of 404.
+    if isinstance(e, GitOperationError):
+        logger.error("State store error", error=str(e), exc_info=True)
+        return jsonify(
+            {
+                "success": False,
+                "message": f"State store unavailable: {e}",
+                "error_type": type(e).__name__,
+            }
+        ), 500
 
     if isinstance(e, HTTPException):
         return jsonify(
--- a/orchestrator/api.py
+++ b/orchestrator/api.py
@@ -147,6 +147,26 @@ def handle_unhandled_exception(e: Exception) -> tuple[Response, int]:
     """Return JSON for all unhandled exceptions."""
     from werkzeug.exceptions import HTTPException
 
+    # Surface state-store infrastructure failures (e.g. ``git worktree``
+    # contention) with the actual error so operators can diagnose without
+    # grepping logs.  Routes used to swallow ``GitOperationError`` as
+    # ``PipelineNotFoundError`` and return 404, masking a recoverable
+    # wedge as "missing pipeline" (#2167).
+    try:
+        from state_store import StateStoreError
+
+        if isinstance(e, StateStoreError):
+            logger.error("State store error", error=str(e), exc_info=True)
+            return jsonify(
+                {
+                    "success": False,
+                    "message": f"State store unavailable: {e}",
+                    "error_type": type(e).__name__,
+                }
+            ), 500
+    except ImportError:  # pragma: no cover — defensive
+        pass
+
     if isinstance(e, HTTPException):
         return jsonify(
             {
--- a/orchestrator/routes/health.py
+++ b/orchestrator/routes/health.py
@@ -35,7 +35,7 @@ def _probe_state_store() -> tuple[bool, str]:
     """Probe whether the state-store worktree is loadable.
 
     Walks the configured ``EGG_REPO_PATH`` (or each child repo in
-    multi-repo mode) and calls ``_ensure_worktree`` on each.  Reports
+    multi-repo mode) and accesses ``store.worktree`` on each.  Reports
     degraded if any probe raises — typically ``GitOperationError`` from
     ``git worktree add`` contention or a stale ``prunable`` admin dir
     (#2167).  Returns the actual error message so the operator sees it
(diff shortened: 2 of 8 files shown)


jwbron:
Filed the deferred follow-ups from the review:

- #2 (concurrent self-heal one-shot error) → #2177
- #6 (`_collect_all_pipelines` silent skip) → #2175
- #7 (probe returns on first failing repo) → #2176

Also filed #2178 covering the broader pattern: feedback-addressing agent shouldn't promise follow-ups it doesn't file.

PR #2240, opened 2026-04-29
Fix #2239: recommend Monitor over Bash for wait-status streaming

### LINKED ISSUE #2239: skills/sdlc: monitor wait-status via Monitor tool, not Bash, for per-event LLM wakes
## Problem

The SDLC skill (`skills/sdlc/SKILL.md`) prescribes invoking `skills/sdlc/bin/wait-status <task_id> --since "<last_cursor>"` via **Bash** for every quiet stretch (Phase 3 / Phase S5 / Critical Rules). In practice this doesn't deliver the per-event wake semantics the skill assumes ("Read each emitted JSON line as it arrives" — line 4 of the Phase 3 monitor section).

`wait-status` is designed to emit one JSON-line per pipeline-relevant event over the lifetime of a single CLI invocation. Both Bash modes break that:

- **Foreground Bash**: the LLM is blocked from the call until the bash exits. Multiple events emitted within the 10-minute cap are batched together at exit, not handled "as they arrive." If the gate-creating `decision.created` event lands 30 seconds in but a CONSENSUS_CONFIRMED lands 8 minutes in, the LLM doesn't see the gate until the second event flushes the buffer.
- **Background Bash** (`run_in_background: true`, or auto-backgrounded): output goes to a file, and the LLM gets *one* completion notification when the whole CLI exits. New stdout lines mid-run produce no notification — the LLM has to poll the file, which defeats the cursor-threading liveness model.

The harness has a `Monitor` tool whose stdout-line semantics are exactly what the skill wants: each line of stdout becomes a separate notification that wakes the LLM. Same `wait-status` command, same cursor threading, same exit-code contract — but events arrive as they fire.

## Hit on issue #1557 SDLC run

First `wait-status` invocation backgrounded silently (no per-event wakes); a foreground retry would have batched the refine-phase events (3× CONSENSUS_CONFIRMED, phase.completed, decision.created) into one wake at exit, instead of surfacing the gate immediately on `decision.created`. Switching to Monitor produced one notification per event as designed.

## Proposed change

Update `skills/sdlc/SKILL.md` to recommend Monitor as the primary wrapper. Bash stays as a fallback only when Monitor isn't available, with the caveat that events batch at exit.

Specific spots that name Bash explicitly and need updating:

- **Line 315 / 1295** ("Drive the pipeline through one Bash invocation per quiet stretch")
- **Line 319 / 1299** ("invoke the canonical Bash idiom")
- **Line 348 / 1328** ("thread it through `--since` on the next Bash invocation")
- **Line 373 / 1350** ("Do NOT wrap the Bash invocation in an outer for-loop or sleep")
- **Line 990** (MCP Tools Reference table row: "skills/sdlc/bin/wait-status (Bash)")
- **Line 1005** ("Polling protocol: ... uses one Bash invocation")
- **Line 1010** (Critical Rules: "every subsequent quiet stretch uses ... (Bash)")
- **Line 1435** (Short Flow Critical Rules: "For blocking waits use the Bash launcher")

The 10-minute Bash cap + re-invocation guidance becomes irrelevant under Monitor (Monitor's timeout is up to 1 hour, and it streams per-line). Wording for cap-elapsed re-invocation should be moved to the fallback section.

## Out of scope

- The `wait-status` CLI itself is fine — no change needed there.
- The orchestrator's `/status/wait` route contract is fine.
- Cursor threading / trigger allowlist / exit-code contract — unchanged.

This is purely a skill-documentation update so the LLM picks the right harness wrapper for the streaming-stdout pattern.

## Related

- #1932 (closed) added `wait-status` and the host-side event-driven wake — this issue is the documentation follow-up so the skill matches the actual harness affordances.

jwbron:
## Summary

- Update `skills/sdlc/SKILL.md` Phase 3 and Phase S5 to wrap `skills/sdlc/bin/wait-status` in the **Monitor** tool, not Bash. Monitor delivers each stdout JSON-line as its own notification, so the LLM wakes on every emitted event in real time — exactly what the skill's "Read each emitted JSON line as it arrives" model assumes. Foreground Bash blocks until exit and batches every event in the 10-min window into one wake; background Bash sends a single completion notification. Either way, a `decision.created` that lands 30 s into a wait won't surface until the next event flushes the buffer.
- Bash stays as a documented fallback with the batches-at-exit caveat. The MCP Tools Reference table row, the Polling protocol prose, the Critical Rules, and the Short Flow Critical Rules all flip to Monitor-as-primary.
- Add an `allowed-tools` line to the skill frontmatter (matching an existing internal skill's convention) listing Monitor, `Bash(skills/sdlc/bin/wait-status:*)`, the gh/git commands the skill prescribes, AskUserQuestion, Read/Grep, and the MCP egg tools — so users running `/sdlc` don't get prompted for the commands the skill is designed to invoke.

## Out of scope

- The `wait-status` CLI itself, the orchestrator's `/status/wait` route contract, the cursor format, the trigger allowlist, and the exit-code contract are all unchanged. This is purely a skill-documentation update so the LLM picks the right harness wrapper for the streaming-stdout pattern.

Closes #2239.

## Test plan

- [ ] Visually re-read Phase 3 and Phase S5 — the Monitor invocation block should be the canonical idiom; Bash appears only as a fallback subsection.
- [ ] Visually re-read the MCP Tools Reference, the Polling protocol prose, and both Critical Rules sections — every reference to the wait-status wrapper says Monitor first, Bash second.
- [ ] Re-run `/sdlc` end-to-end on a real pipeline and confirm: (a) the LLM uses Monitor for the blocking wait, (b) per-event wakes land for `decision.created` (HITL gates surface immediately), (c) the cursor threads correctly across re-invocations on Monitor `timeout_ms`.
- [ ] Confirm no permission prompts for the wait-status command when running the skill.

### DIFF
--- a/skills/sdlc/SKILL.md
+++ b/skills/sdlc/SKILL.md
@@ -3,6 +3,7 @@ name: sdlc
 description: "Run an egg SDLC pipeline: full lifecycle (default) or lightweight coder+reviewer with --short."
 disable-model-invocation: true
 argument-hint: "[--short] [--qualifier <name>] [JIRA-1234 or issue# or description] [--repo owner/name]"
+allowed-tools: Monitor Bash(skills/sdlc/bin/wait-status:*) Bash(gh issue view:*) Bash(gh issue list:*) Bash(gh pr list:*) Bash(gh pr view:*) Bash(git remote:*) Bash(git -C *:remote:*) Bash(git log:*) Bash(git show:*) Bash(git ls-tree:*) Read Grep AskUserQuestion mcp__egg__submit_task mcp__egg__get_status mcp__egg__provide_input mcp__egg__list_tasks mcp__egg__cancel_task mcp__egg__check_health mcp__egg__list_containers mcp__egg__get_container_logs mcp__egg__send_message mcp__egg__get_consensus_status mcp__egg__get_phase mcp__egg__get_pipeline_snapshot mcp__egg__get_contract mcp__egg__list_checkpoints mcp__egg__search_checkpoints
 ---
 
 # SDLC Pipeline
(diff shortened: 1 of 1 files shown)


jwbron:
Thanks for the re-review and approval. Quick status on the three non-blocking observations:

**#1 PR description still references an internal skill-convention name.** Tried `gh pr edit 2240 --body-file ...` and the gateway refused (`Edit denied: PR #2240 is not owned by james-in-a-box or configured user (author: jwbron)`) — egg can't edit PRs authored by someone else. Posting the corrected body text here so the author (or a maintainer) can paste it in:

> Add an `allowed-tools` line to the skill frontmatter (per the [Claude Code plugin skills frontmatter docs](https://code.claude.com/docs/en/plugins/skills#frontmatter)) listing Monitor, `Bash(skills/sdlc/bin/wait-status:*)`, the gh/git commands the skill prescribes, AskUserQuestion, and the MCP egg tools — so users running `/sdlc` don't get prompted for the commands the skill is designed to invoke.

(The only delta vs. the current third bullet is dropping the internal skill-convention reference and replacing it with the linked public-docs pointer.)

**#2 `(auto-stopped)` recovery could loop in pathological cases.** Acknowledged. Leaving the current single-row guidance as-is for this PR — the cursor-advance argument you laid out (bursts are bounded by phase boundaries, so the second invocation naturally sees only fresh events) is sound, and the SKILL.md edit budget is already at its scope ceiling. If we see a real loop in production I'll add the "fall back to one-shot `get_status` polling on the second consecutive auto-stop" sentence in a follow-up.

**#3 Trailing `:*` form is permissive.** Confirming this isn't a regression — the existing `Bash(git remote:*)` rule already permits `git remote add` / `git remote remove`, and tightening to `Bash(git -C * remote get-url:*)` would mean a permission prompt the next time the skill needs a different read-only `git remote` verb (e.g. `git remote -v` in a debug path). Defense-in-depth tradeoff falls on the side of the existing scope; happy to revisit if a stricter rule becomes worthwhile.

— Authored by egg

Issue #2224, opened 2026-04-28
Defense-in-depth guardrails for pipeline-branch contamination (Phase 4 of #2222)


jwbron:
Follow-up to #2222. The fix PR for #2222 addresses the root cause (the `_build_rebase_cmd` unsafe fallback + the `_rebase_pipeline_branch_onto_base` step-4 skip) and the precondition (stale-branch reuse on re-submit). This issue tracks three independent guardrails that increase the cost of any future regression in this area.

Each item below is independent — split into separate PRs as appropriate.

## (1) Gateway policy: reject agent `git rebase` against `origin/<base_branch>`

The pipeline branch should track main only via the orchestrator's controlled rebase. Add a policy in `gateway/git_client.py` (around the existing rebase allowlist near line 615) that rejects `git rebase` whose upstream argument resolves to `origin/<base_branch>` (typically `origin/main`) when invoked from an agent session. Orchestrator-side rebases bypass this policy.

**Why**: an agent reaching for `git rebase origin/main` (intentionally or via a "resolve conflicts" intuition) reproduces the exact contamination shape from #2222 even with the orchestrator-side fixes in place.

**How to apply**: rejection should produce a clear error in the agent's transcript pointing at `_rebase_pipeline_branch_onto_base` as the right path.

## (2) End-of-pipeline rebase against current main

Confirmed during #2222 investigation: nothing between branch cut and PR open refreshes against `origin/main`. `_rebase_pipeline_branch_onto_base` runs at the start of each phase iteration, but if main advances *during* the PR phase's own work, the resulting PR is behind. `_auto_create_pr` itself does no fetch/rebase.

Add a controlled rebase against current `origin/main` immediately before `_auto_create_pr` calls `gateway.create_pr` so the resulting PR has a clean linear diff. Use the same `--onto origin/<branch> origin/<base>` form to avoid the contamination shape.

**Note**: this is a user-visible behavior change for downstream reviewers — the PR head SHA at open time will reflect the rebase. Worth its own PR description.

## (3) Audit logging on branch divergence

When the pipeline-watcher observes the pipeline branch is more than N commits ahead of `origin/<base>` AND those ahead-commits contain merged-PR signatures (e.g., `(#NNNN)` PR refs in the subject, or a `Co-Authored-By` footer for an account other than the agent committers), emit an `OVERSEER_ALERT` with the offending SHAs.

[...]

Issue #2222, opened 2026-04-28
Investigate why pipeline branch ate main commits + carried stale prior-run history


jwbron:
## Problem

Pipeline `issue-2137`'s branch (`egg/issue-2137`) shipped a borked diff: 65 commits ahead of `origin/main` and 41 commits on `origin/main` not in the branch. The branch contained two distinct surprises:

1. **Pre-pipeline stale commits**: 4 commits at the bottom of the branch (`908a65559 Initialize SDLC contract`, `15e411b94 refine: add analysis for #2137`, `f9de3ad56 refine: revise #2137 analysis per reviewer_refine NACK`, `bf8f9e665 Persist statefiles after refine phase`) that pre-date the *current* pipeline's first commit (`92b8af1ed`). These appear to be remnants of a **prior failed `/sdlc` invocation** on the same issue/branch.

2. **Absorbed main commits**: ~30 commits from `main` got rebased *into* the branch's linear history (`fcfa022d1`, `fcbff05bb`, `ee94d2f60`, …, `335069f7f`). These are merged PRs from main (`#2138`, `#2150`, `#2152`, etc.) that should have been parents of the branch via a regular `git merge` or rebase target — instead they appear as **first-class commits in the branch's linear log**, consuming SHAs and showing up in `git log origin/main..egg/issue-2137`.

Net effect: the GitHub PR (#2220) showed `mergeStateStatus: DIRTY / mergeable: CONFLICTING` after the pipeline completed. The diff against current main was 65+ extra commits with multiple modify/delete and content conflicts in `.egg-state/` and `orchestrator/env_config.py`.

The branch was force-rebased onto `b1325eede` (PR #2179, the head when the pipeline started) to clean it up. That made the branch's content linear and sensible, but the *root cause* of the original mess was not addressed.

## Reproduction

`/sdlc https://github.com/jwbron/egg/issues/2137` on 2026-04-28. Branch `egg/issue-2137` opened by `submit_task`. Pipeline ran for ~5h 40min through refine → plan → implement → pr phases. Manual unsticks via `restart_agent`/`complete_phase`/`advance_phase` for unrelated bugs (#2196, #2200, #2203, #2204, #2205, #2210, #2219). PR #2220 created at 21:54:19 UTC.

`gh pr view 2220 --json mergeable,mergeStateStatus` returned `\"CONFLICTING\", \"DIRTY\"`.

[...]

PR #2282, opened 2026-04-29
Part of #2224 (1/3): gateway blocks bare `git rebase origin/main` from agents

jwbron:
## Summary

Defense-in-depth guardrail (Phase 4, item 1 of 3) for #2224, follow-on to #2222.

The pipeline branch is rebased onto the base branch only via the orchestrator's controlled rebase in `_rebase_pipeline_branch_onto_base`, which uses the safe `--onto <new_base> <old_base> <branch>` form on a subprocess that does not route through `/api/v1/git/execute`. An agent reaching for the *bare* form `git rebase origin/main` (intentionally or via a "resolve conflicts" intuition) reproduces the contamination shape from #2222 even with the orchestrator-side fixes in place.

## Changes

- **gateway/gateway.py**: New guard in `git_execute` that rejects bare-form `git rebase origin/main` (or bare `main`) from sessions with an `assigned_branch` (pipeline sessions). The `--onto` form is allowed: it is the canonical safe shape used by the stacked-PR healer (`gateway_client.rebase_onto`) which legitimately routes through this endpoint. The `--onto=X` (equals) form is also recognized.
- **gateway/tests/test_pipeline_enforcement.py**: New `TestBareRebaseAgainstBaseBlocking` class with 8 tests covering bare-form rejection (`origin/main`, `main`, with branch arg), `--onto` allowance (separated and equals form), bare rebase against non-main branches (allowed), `--continue` (allowed), and non-pipeline sessions (allowed).

## Why

The denial message points at `_rebase_pipeline_branch_onto_base` so the agent transcript explains the recovery path. Orchestrator-side rebases (subprocess, not routed through this endpoint) bypass this policy by construction.

## Scope

This is **only** PR 1 of 3 for issue #2224. The umbrella issue tracks two more independent guardrails:

- (2) End-of-pipeline rebase against current main, before `_auto_create_pr` opens the PR.
- (3) Pipeline-watcher `OVERSEER_ALERT` on branch divergence with merged-PR signatures.

## Test plan

- [x] New unit tests pass (8/8).
- [x] Existing `gateway/tests/test_pipeline_enforcement.py` passes (23/23 total).
- [x] `ruff check` clean.
- [ ] CI green.


jwbron:
Verified: #2222 regression test (`test_rebase_does_not_contaminate_when_base_fetch_silently_failed`) still passes on this branch.

PR #2290, opened 2026-04-29
Part of #2224 (3/3): OVERSEER_ALERT on pipeline-branch divergence with merged-PR signatures

jwbron:
## Summary

Defense-in-depth guardrail (Phase 4, item 3 of 3) for #2224, follow-on to #2222.

When a pipeline branch is materially ahead of base AND the ahead-commits contain merged-PR subject signatures (\`(#NNNN)\`), the pipeline-watcher emits an \`OVERSEER_ALERT\` so operators see the contamination shape from #2222 in real time rather than only at PR open.

## Why

A real pipeline branch grows by refine/plan/implement/state-file commits authored by agents — none of those carry a \`(#NNNN)\` suffix in the subject. When that signature appears alongside material divergence, the branch has absorbed merged-main commits — the exact failure mode that produced #2220 (the contaminated PR for #2137). The signature heuristic is intentionally cheap and false-positive-tolerant; per the issue, we'd "rather over-alert than miss another contaminated PR."

## Changes

- **orchestrator/routes/pipelines.py**:
  - New \`BRANCH_DIVERGENCE_THRESHOLD = 20\` and \`_BRANCH_DIVERGENCE_PR_RE = re.compile(r\"\\(#\\d+\\)\")\`.
  - New helper \`_check_branch_divergence_for_alert(...)\` runs \`git rev-list --count\` then \`git log --pretty=format:%H%x09%s\` and filters to subjects matching the PR-signature regex. All errors logged-and-swallowed.
  - New helper \`_publish_branch_divergence_alert(...)\` mirrors \`_publish_consensus_timeout_alert\` (#2264): publishes \`OVERSEER_ALERT\` with offending SHAs in metadata, ties back to #2222 in the body.
  - Wired into \`_health_monitor_poll\` (every 30s tick) with a per-pipeline \`divergence_alerted_shas\` dedupe set so each offending commit produces at most one alert per pipeline run.

- **orchestrator/tests/test_branch_divergence_alert.py**: 13 tests covering: not-ahead, at-threshold (no alert), above-threshold without signatures (no alert), above-threshold with signatures (alert), all best-effort error paths (rev-list/log fails, subprocess timeout), branch == base no-op, threshold override, alert content + metadata, message-store-unavailable graceful skip, message-store exception swallow, offender-list truncation in body.

## Scope

This is **only** PR 3 of 3 for issue #2224. PR 1 (gateway-side bare-rebase block) is in #2282; PR 2 (end-of-pipeline rebase) is next.

## Test plan

[...]

PR #2291, opened 2026-04-29
Part of #2224 (2/3): rebase pipeline branch against current base_branch before PR open

jwbron:
## Summary

Defense-in-depth guardrail (Phase 4, item 2 of 3) for #2224, follow-on to #2222.

Closes the gap where `base_branch` advances *during* the PR phase: before this fix, `_auto_create_pr` opened the PR against whatever tip `origin/<pipeline_branch>` had — which could be N commits behind current `origin/<base_branch>`. `_rebase_pipeline_branch_onto_base` runs at the start of each phase iteration to clean up stale branch state on resume; nothing between branch-cut and PR-open refreshes against `origin/<base_branch>`.

`base_branch` is read-only here. The helper only rewrites `pipeline_branch`; no commits are ever pushed to `base_branch`, even when it happens to be `main`.

## Changes

- **orchestrator/routes/pipelines.py** — new helper `_refresh_pipeline_branch_against_current_base`:
  1. Fetches fresh refs.
  2. No-ops when `origin/<pipeline_branch>` is already current with `origin/<base_branch>`.
  3. Computes the merge-base and rebases via the safe 3-arg `--onto` form (`--onto <new_base> <merge_base>`) — the same contamination-safe shape `_rebase_pipeline_branch_onto_base` uses post-#2222.
  4. Force-pushes the rebased tip of `pipeline_branch` so the open PR's head SHA reflects current `base_branch`.

  Wired into `_auto_create_pr` immediately before `gateway.create_pr`. Best-effort: any failure (rebase conflict, push reject, transient gateway error) restores the worktree to `origin/<pipeline_branch>`, logs at WARNING, and returns False — the caller still opens the PR against the un-rebased tip. This is intentional: a merge conflict at PR-open time is better surfaced to the human reviewer than swallowed by failing the whole pipeline.

- **orchestrator/tests/test_refresh_pipeline_branch_at_pr_open.py** — 13 tests covering: empty/equal-branch no-ops, fetch failure, missing origin refs, already-current no-op, merge-base failure, reset failure, rebase conflict (abort + restore), push failure (restore + return False), full success path (rebase + push + re-fetch), and an explicit assertion that the rebase argv uses the safe `--onto <new_base> <merge_base>` form (so a refactor that drops `--onto` is caught at unit-test time).

## User-visible behavior change

The PR head SHA at open time will now reflect the rebase when `base_branch` advanced during the PR phase. Reviewers will see a head SHA that didn't exist locally during the implement phase. The diff against current `base_branch` is cleaner; the change to the resulting PR is purely "less divergence", which is the whole point.

## Scope

This is **only** PR 2 of 3 for issue #2224. PR 1 (gateway-side bare-rebase block) is in #2282; PR 3 (divergence `OVERSEER_ALERT`) is in #2290. Each is independent.

## Test plan

- [x] New unit tests pass (13/13).
- [x] No regression in adjacent rebase tests (`test_rebase_pipeline_branch.py`, `test_reconcile_and_push_pr_branch.py`).
- [x] `ruff check` clean.
- [ ] CI green.
- [ ] Manual: verify a real pipeline run where `base_branch` advances mid-PR-phase produces a PR with head SHA matching the rebased tip.


jwbron:
Verified: #2222 regression test (`test_rebase_does_not_contaminate_when_base_fetch_silently_failed`) still passes on this branch.

Issue #2512, opened 2026-05-07
restart_phase fails to recover when slice integration branch already exists with non-ancestor commits from prior run


jwbron:
## Observation

Pipeline `issue-2474` was cancelled (`cancel_task cleanup=false`) after slice-1 implement-phase agents accumulated work but never reached BRC consensus. The slice-1 integration branch `egg/issue-2474/slice-1` on origin contained four commits from the prior run:

```
0c5c66978 slice-1 (tester absorbs gateway-blocked work)
e2f760136 slice-1 (coder followups)
6edb6a448 slice-1 (coder portion)
73560c0c5 slice-1/task-1-2 (tester scope)
8a76c30d3 Persist statefiles after plan phase  ← parent-branch tip
```

The operator then called `restart_phase(task_id="issue-2474", phase="implement")`. The MCP call returned `restarted: true, agents_restarted: [...8 roles...]`. But within ~63 seconds the pipeline transitioned to `pipeline.failed`. Orchestrator logs:

```
01:04:02 WARNING gateway_client: Failed to create slice integration branch
  pipeline_id=issue-2474
  integration_branch=egg/issue-2474/slice-1
  parent_branch=egg/issue-2474/work
  parent_sha=8a76c30d3cc1628601432b91c76bb0f118b91bcc
  error="Push failed: To https://github.com/jwbron/egg.git ! [rejected] 8a..."

01:04:02 ERROR pipelines: Slice integration branch creation failed; marking slice failed (agents not spawned)
01:05:02 WARNING pipelines: Slice cascade fired failed_slice=slice-1 blocked=['slice-2', 'slice-3', 'slice-4', 'slice-5']
01:05:02 ERROR pipelines: Phase failed exit_code=1
```

The orchestrator tried to (re-)create the slice-1 integration branch by pushing the parent-branch tip (`8a76c30d3`) to `refs/heads/egg/issue-2474/slice-1`. That push is NOT a fast-forward over the existing slice-1 commits, so GitHub rejected it. The slice was marked failed, the slice cascade fired, and downstream slices 2–5 were blocked.

## Why it matters

- `decision-3` (the "phase failed → retry/accept/abort" gate) advertises *"Committed work is preserved on the per-role branch — 'Retry phase' restarts with artifacts intact"*. That's not true in practice for slice-based implement phases — the integration-branch push rejection happens before any agent is spawned, so even the artifacts that ARE preserved on per-role branches aren't reachable from the new run.
[...]

Issue #2409, opened 2026-05-05
Slice-scoped consensus trackers can't reconstruct from message store after orchestrator restart


jwbron:
## Problem

Per-slice consensus trackers are registered in-memory under the
nested key ``{pipeline_id}/{slice_id}`` (see
``orchestrator.peer_consensus._tracker_key``). When the orchestrator
restarts mid-pipeline, the in-memory ``_trackers`` dict is empty;
``handle_consensus_confirmed_signal`` falls back to
``reconstruct_tracker_from_messages(pipeline_id, graph)`` to replay
state from the message bus.

That replay is only safe for **pipeline-level** trackers. Messages
in the store are keyed by the bare ``pipeline_id`` — there is no
``slice_id`` field to filter on — so a per-slice replay would mingle
every slice's CONSENSUS_* messages and reach a false consensus.
After #2403's fix, the confirmed handler explicitly skips
reconstruction when ``slice_id`` is supplied:

\`\`\`python
if slice_id is None:
    graph = get_review_graph_for_phase(_phase, repo=_repo)
    tracker = reconstruct_tracker_from_messages(pipeline_id, graph)
\`\`\`

So a slice agent sending ``consensus_confirmed`` after an orchestrator
restart gets a 404 and the slice can't auto-recover.

## Why this is latent (not actively failing today)

- Live operation is fine: the run loop creates the tracker once when
  it spawns the slice's BRC team, and CONSENSUS_* routes there for the
  rest of the slice's lifetime.
- Pre-#2403, slice routing was end-to-end broken (\`/\` in
  ``EGG_PIPELINE_ID`` 4xx'd every agent → orchestrator round-trip), so
  this gap was strictly worse but masked by a louder upstream failure.
- Mid-implement orchestrator restarts aren't a routine operator
  workflow today.

## Fix sketch

Two reasonable shapes:

**A. Tag messages with slice_id.** Add a ``slice_id`` column / field
to ``message_store.Message`` and have ``store.add_message`` write it
when the signal handler received one. ``reconstruct_tracker_from_messages``
gets a ``slice_id`` parameter and filters the replay window. Backwards-
compatible: ``slice_id=None`` replays bare-pipeline messages exactly
as today.

**B. Persist slice trackers explicitly.** Snapshot tracker state to
disk (the same place pipeline state lives) on every CONSENSUS_*
mutation, restore on orchestrator startup. Heavier but generalises
to non-message-store-derived state too.

[...]

Issue #1473, opened 2026-03-31
BRC protocol should require commit+push before consensus confirmation


jwbron:
## Summary

The BRC (Broadcast-Review-Confirm) consensus protocol currently allows agents to confirm consensus without verifying that their work has been committed and pushed. This creates a gap where:

1. An agent writes code and proposes via `CONSENSUS_PROPOSE`
2. Reviewers ACK the proposal (potentially reviewing code that isn't yet on the branch)
3. Agent confirms via `CONSENSUS_CONFIRMED`
4. Agent exits — and `auto_commit_worktree()` catches uncommitted work as a WIP commit
5. The WIP auto-commit lands in the PR **without BRC review**

## Observed in Pipeline

Pipeline `issue-1470` (PR #1472) produced a `WIP: auto-commit uncommitted work` commit (`5f7d39f`) from a failsafe push container. This commit modified test files (`test_agent_restrictions_enforce.py`, `test_agent_restrictions_filter.py`) that were never reviewed through BRC, and likely caused lint check failures.

## Proposed Change

Update the BRC consensus protocol to enforce two invariants:

### 1. Producers must commit and push before proposing

Agents with write access (coder, tester, documenter, autofixer, conflict_resolver) must have all their work committed and pushed to the shared branch **before** sending `CONSENSUS_PROPOSE`. The proposal should reference the actual commit SHA(s) on the remote branch, not just a summary of planned work.

This could be enforced by:
- Having the `CONSENSUS_PROPOSE` signal include a `commit_sha` field (the HEAD SHA on the remote branch after push)
- The orchestrator validates that the SHA exists on the remote branch before accepting the proposal
- If the agent hasn't pushed, the proposal is rejected with a message asking the agent to push first

### 2. Reviewers should only review committed work

Reviewer agents (reviewer_code, reviewer_contract) should verify that the code they're reviewing is actually on the shared branch. When reviewing a proposal:
- Pull latest from the remote branch
- Verify the producer's claimed `commit_sha` matches what's on the branch
- Review the actual diff from the commit, not just the proposal summary

This ensures reviewers are reviewing the same code that will land in the PR.

### 3. Auto-commit should not introduce unreviewed work

If `auto_commit_worktree()` fires for an agent that already confirmed consensus, the auto-committed changes represent **unreviewed work** that bypassed BRC. Options:
- Flag auto-commits that happen post-confirmation as warnings in the pipeline status
- Prevent auto-committed work from being pushed (only preserve locally for debugging)
[...]

Issue #1471, opened 2026-03-31
Documenter stuck in heartbeat loop after consensus re-review; overseer doesn't detect or nudge


jwbron:
## Summary

After a coder v3 re-proposal triggered `CONSENSUS_RE_REVIEW`, the documenter agent — which had already confirmed in round 1 — entered a heartbeat-only loop and never re-confirmed. The pipeline stalled for 15+ minutes with 4/5 agents confirmed. The overseer did not detect this as a problem and did not nudge. Manual intervention via `send_message` was required to unblock.

## Timeline (pipeline-d9ac3268, jwbron/testing)

| Time | Event |
|------|-------|
| 20:53:35 | Documenter confirms (round 1) |
| 21:08:15 | Coder submits v2 proposal → `CONSENSUS_RE_REVIEW` sent to all agents |
| 21:09:08 | Coder withdraws v2, re-proposes v3 |
| 21:09:23 | reviewer_code and reviewer_contract ACK + confirm v3 |
| 21:14:04 | tester ACKs + confirms v3 |
| 21:14:06 | coder confirms v3 |
| 21:14–21:23 | **Documenter stuck** — receives all messages (including RE_REVIEW and v3 proposal) but only sends heartbeats, never re-confirms |
| 21:23:10 | Manual nudge sent via `send_message`: "You need to re-confirm via `egg-orch consensus confirmed`" |
| 21:23:15 | Documenter confirms (5 seconds after nudge) |

## What the overseer did

- Detected heartbeat/progress stalls for several agents earlier in the pipeline and correctly resolved them as stale
- Tracked coder as "blocked" waiting for tester re-ACK (which resolved naturally)
- At the time the documenter was stuck, the overseer reported: "Pipeline stable — agents in BRC stay alive phase post-consensus. This is expected behavior." with 0 alerts
- **Did not detect** that consensus was incomplete (4/5 confirmed) or that the documenter was the blocking agent
- **Did not nudge** the documenter

### 1. Documenter doesn't handle re-review in stay-alive loop

After confirming in round 1, the documenter enters a heartbeat + message-poll loop. When the `CONSENSUS_RE_REVIEW` arrives, the documenter sees it in its poll results but doesn't act on it — it just continues heartbeating. The stay-alive loop likely lacks logic to:
- Detect `CONSENSUS_RE_REVIEW` messages
- Re-enter the confirmation flow when consensus is reset

### 2. Overseer doesn't track consensus completeness

[...]

PR #2542, opened 2026-05-07
Fix #2535: stop slice-N from inheriting slice-(N-1) consensus, drop gateway import

### LINKED ISSUE #2535: Slice-2 spawn-to-fail in 4 seconds: stale BRC consensus tracker declares 'consensus reached' at elapsed_seconds=0.0; rebase_onto crashes on missing module
## Observation

After slice-1 of pipeline \`issue-2474-v2\` reached BRC consensus and completed at 04:03:41 UTC, the orchestrator transitioned to spawning slice-2 (\"Promote ScriptedProvider\"). Slice-2 spawned + failed in **4 seconds** with no agent work attempted. Two distinct bugs in the spawn → consensus → PR-creation pipeline:

\`\`\`
04:04:32 ERROR  orchestrator.gateway_client: rebase_onto: gateway/git_client module unavailable pipeline_id=issue-2474-v2 [/app/gateway_client.py:1426]
04:04:44 INFO   orchestrator.kubernetes_spawner: 8 spawn_attempt outcomes — all success (coder, tester, documenter, 5 reviewers)
04:04:45 INFO   orchestrator.pipelines: Consensus reached, stopping containers pipeline_id=issue-2474-v2 elapsed_seconds=0.0 has_failures=False [/app/routes/pipelines.py:13291]
04:04:49 ERROR  orchestrator.pipelines: Slice PR creation failed pipeline_id=issue-2474-v2 slice_id=slice-2 error=\"Failed to create PR: pull request create failed: GraphQL: No commits between ...\" [/app/routes/pipelines.py:12500]
04:05:49 WARNING orchestrator.pipelines: Slice cascade fired failed_slice=slice-2 blocked=['slice-3', 'slice-4', 'slice-5']
04:05:49 ERROR  orchestrator.pipelines: Phase failed pipeline_id=issue-2474-v2 phase=implement exit_code=1
\`\`\`

### Bug A — \`rebase_onto: gateway/git_client module unavailable\`

\`gateway_client.py:1426\` attempts a \`rebase_onto\` operation as part of slice integration branch reconciliation (the path added in the #2512 fix), and the import / lookup for \`gateway/git_client\` fails. Logged at ERROR but doesn't block spawn — the orchestrator proceeds with spawning slice-2 agents anyway. May be a deployment-packaging issue (module not shipped), or a path/import bug in the new code.

This is a regression / incomplete fix from #2512: the new branch-reconciliation path depends on a module that isn't available in the deployed orchestrator image. Even though it doesn't immediately fail spawn, it's a broken code path.

### Bug B — Stale BRC consensus tracker declares \"consensus reached\" at elapsed_seconds=0.0

This is the actual cause of the slice-2 failure. \`routes/pipelines.py:13291\` fires:

\`\`\`
Consensus reached, stopping containers pipeline_id=issue-2474-v2 elapsed_seconds=0.0 has_failures=False
\`\`\`

at exactly the same second slice-2 agents finished spawning. There is no possible way real BRC consensus reached in 0.0 seconds — agents need time to load context, propose, ACK, NACK, confirm. The orchestrator's consensus tracker was reading **stale slice-1 state** (where 8 agents had just confirmed legitimately) and applying it to slice-2's empty BRC matrix (which has the same agent roles but new container IDs). Slice-2's brand new containers were immediately stopped before any work happened. Slice-2's branch (\`egg/issue-2474-v2/slice-2\`) ended up with zero commits.

Then 4 seconds later, the orchestrator tried to create slice-2's PR and failed with \`pull request create failed: GraphQL: No commits between ...\` — exactly what you'd expect from an empty branch.

This matches and strengthens #2502 (snapshot returns stale prev-phase consensus): the consensus tracker doesn't reset on slice / phase boundary, so when a new slice spawns with the same agent roles, the tracker sees them as \"already confirmed\" from the prior slice's run.

### Why it matters

- Slice-2 (and by cascade, slices 3, 4, 5) was guaranteed to fail. The user's pipeline cannot deliver more than one slice per pipeline run until this is fixed.
- The PR-creation failure log message *\"No commits between ...\"* is the visible symptom; the upstream cause (false consensus on a brand new slice) is invisible unless you trace back through the orchestrator log.
- The #2512 fix path (\`rebase_onto\`) is silently broken in deployment — needs to be exercised in CI to catch the missing-module issue before it lands in prod.

### Suggested fixes

**For Bug A** (`rebase_onto` missing module):

1. Audit the import / module registration in \`gateway_client.py:1426\`. Likely \`gateway/git_client.py\` should be a sibling module or properly packaged. Add an import-time sanity check at orchestrator startup that exercises this code path so deployment-time misses are caught fast.

2. The current \`rebase_onto\` path falling back silently means deployments with the missing module appear healthy until they hit a slice integration branch reconciliation. Make the failure loud (refuse to start, or transition pipeline to FAILED with a clear error).

**For Bug B** (stale consensus tracker):

3. Reset the BRC consensus tracker on slice boundary. Specifically: when transitioning from slice N to slice N+1, clear \`concurrent.consensus.agents\` and re-initialize for the new slice's spawned agents. The agent roles may match across slices, but the proposals / ACKs / CONFIRMs are per-slice and per-container-id and should not carry over.

4. Add a defensive guard at \`routes/pipelines.py:13291\`: before declaring \"consensus reached, stopping containers\", verify \`elapsed_seconds >= some-threshold\` (e.g. 10s or based on \`brc_consensus_progress_gate_seconds\`). Zero-elapsed consensus is impossible by physics — refuse to honor it.

5. Also tie into #2502: the snapshot's \`concurrent.consensus\` field has the same staleness problem. A unified fix that resets on slice / phase boundary addresses both.

### Visible blast radius

For pipeline \`issue-2474-v2\`:
- Slice-1: completed legitimately (3 commits on \`egg/issue-2474-v2/slice-1\`, PR [#2533](https://github.com/jwbron/egg/pull/2533))
- Slice-2: failed instantly (this bug)
- Slices 3–5: cascade-failed (blocked by slice-2)

For any future multi-slice pipeline that uses the same agent roles across slices: the second slice will hit the same issue.

Found while monitoring pipeline \`issue-2474-v2\` (issue #2474). Related: #2502 (stale concurrent.consensus on phase transition), #2512 (slice integration branch creation — partially fixed).

jwbron:
## Summary

Closes #2535. Two bugs that together caused issue-2474-v2 slice-2 to spawn-and-fail in four seconds:

- **Bug A — `gateway/git_client module unavailable`** — `orchestrator/Dockerfile` does not COPY `gateway/`, so `from gateway.git_client import build_rebase_onto_args` (added by #2512 at `orchestrator/gateway_client.py:1424`) always raises `ImportError` in the deployed image. Slice integration branch reconciliation silently fails. Inlined the canonical argv builder as `_build_rebase_onto_args` in the orchestrator package; the gateway server's `/git` endpoint remains the authoritative allowlist boundary.

- **Bug B — slice-2 false consensus at `elapsed_seconds=0.0`** — `ConcurrentPhaseExecutor.check_consensus()` had two slice-unaware fallback paths (tracker reconstruction and the message-bus subset check). The per-slice tracker registry is already keyed by `{pipeline_id}/{slice_id}` and `spawn_all` registers a fresh empty tracker per slice, but when that tracker is empty both fallbacks scanned messages under the bare `pipeline_id` — picking up slice-1's CONSENSUS_CONFIRMED messages whose `from_role` set matches slice-2's roster (coder/tester/documenter + 5 reviewers re-spawn per slice) and declaring consensus instantly. Gated both fallbacks (plus the matching path in `handle_consensus_confirmed_signal`) on `slice_id is None`. The in-memory per-slice tracker is now the authoritative source for slice work; an empty fresh tracker correctly returns `is_complete=False`.

The same root cause — message-bus probes scoped only to the bare `pipeline_id` — also showed up in `_existing_confirmed_for_role` (the idempotency probe in the same handler). That made slice-N's first per-role CONFIRMED look "already confirmed" because slice-(N-1)'s same-role CONFIRMED was still in the bus, suppressing the new bus write and the #1473 consensus-confirmed marker. Fixed in the review-feedback follow-up commit by tagging CONSENSUS_CONFIRMED messages with `metadata["slice_id"]` and filtering the probe by it. Pipeline-scoped (no `slice_id`) callers see only pipeline-scoped messages, preserving the legacy non-slice path exactly.

### Out of scope (tracked separately)

The same slice-unaware `get_peer_consensus_tracker(pipeline_id)` / `reconstruct_tracker_from_messages(pipeline_id, ...)` pattern still exists in `orchestrator/kubernetes_monitor.py:915,934`, `startup_reconciliation.py:308,312`, `routes/pipelines.py:4166,4183` (status display), and the tier-1 `incomplete_consensus_stall.py` / `consensus_stall.py` health checks. None of those are consensus-determining (they observe rather than decide), so the worst case is bogus observability. Tracked under #2409, which already scopes the slice-aware reconstruction work and the `slice_id` field on `Message`.

## Test plan

- [x] `make lint` clean
- [x] `pytest orchestrator/tests/test_check_consensus_slice_isolation.py orchestrator/tests/test_gateway_client_rebase_onto.py orchestrator/tests/test_concurrent_executor.py orchestrator/tests/test_consensus.py orchestrator/tests/test_consensus_polling.py orchestrator/tests/test_consensus_confirmed_idempotent.py orchestrator/tests/test_slice_signal_routing.py` — 117 passed (95 from the original PR + 3 new slice-idempotency regressions + 19 from the slice signal routing suite that exercises the new `slice_id` kwarg)
- [x] New regression `test_check_consensus_slice_isolation.py` covers both directions: a fresh slice-2 tracker with stale slice-1 CONFIRMs in the bus must NOT report `is_complete`; a pipeline-scoped (`slice_id=None`) executor still uses the legacy message-bus fallback for #1471/#1615 tracker-loss recovery.
- [x] New regression `TestNoGatewayPackageDependency::test_rebase_onto_succeeds_when_gateway_package_unavailable` patches `__import__` to refuse `gateway*` imports and confirms `rebase_onto` still produces the canonical argv.
- [x] New idempotency regressions in `test_consensus_confirmed_idempotent.py`:
  - `test_slice_two_first_confirmed_not_idempotent_when_slice_one_already_confirmed` — slice-2's first CONFIRMED is NOT short-circuited by slice-1's CONFIRMED in the bus; bus message + #1473 marker are written.
  - `test_slice_idempotency_within_same_slice` — within slice-2, the second CONFIRMED IS deduped.
  - `test_pipeline_scoped_call_ignores_slice_scoped_prior_confirms` — legacy non-slice callers ignore slice-scoped CONFIRMs and don't tag their own writes with a `slice_id`.

### DIFF
--- a/orchestrator/concurrent_executor.py
+++ b/orchestrator/concurrent_executor.py
@@ -636,27 +636,31 @@ class ConcurrentPhaseExecutor:
                 pipeline_id=self.pipeline.id,
                 slice_id=self._slice_id,
             )
-            # Attempt lazy reconstruction from message store
-            # Reconstruction does NOT yet support slice scoping — for
-            # slice-scoped trackers we fall back to fetching the bare
-            # pipeline-id tracker (legacy behaviour). This is acceptable
-            # because reconstruction is a backstop for orchestrator
-            # restarts, not the steady-state path; per-slice trackers
-            # are stateless event consumers and are recreated by the
-            # slice scheduler on the next iteration.
-            try:
-                from peer_consensus import reconstruct_tracker_from_messages
+            # Attempt lazy reconstruction from message store — pipeline-
+            # scoped only. Slice-scoped reconstruction is unsafe today:
+            # CONSENSUS_* messages are persisted under the bare
+            # pipeline_id, so a per-slice replay would mingle siblings'
+            # messages and reach false consensus the moment a fresh
+            # slice spawns roles whose names match an already-confirmed
+            # prior slice. Per-slice trackers are stateless event
+            # consumers and are recreated by the slice scheduler on the
+            # next iteration; the pipeline run loop's empty-tracker
+            # iteration will simply observe is_complete=False, which is
+            # the correct answer for a brand-new slice (#2535).
+            if self._slice_id is None:
+                try:
+                    from peer_consensus import reconstruct_tracker_from_messages
 
-                graph = self._get_review_graph()
-                tracker = reconstruct_tracker_from_messages(self.pipeline.id, graph)
-            except ImportError:
-                pass
-            except Exception as e:
-                logger.warning(
-                    "Tracker reconstruction failed",
-                    error=str(e),
-                    pipeline_id=self.pipeline.id,
-                )
+                    graph = self._get_review_graph()
+                    tracker = reconstruct_tracker_from_messages(self.pipeline.id, graph)
+                except ImportError:
+                    pass
+                except Exception as e:
+                    logger.warning(
+                        "Tracker reconstruction failed",
+                        error=str(e),
+                        pipeline_id=self.pipeline.id,
+                    )
         if tracker:
             result = tracker.evaluate()
             # Message-bus fallback: if reconstruction produced a tracker but
@@ -678,7 +682,9 @@ class ConcurrentPhaseExecutor:
                 # Safety net (#1671): if the tracker has all roles in
                 # confirmed_roles but evaluate() returned False due to stale
                 # NACK edges in the approval matrix (common after NACK →
-                # re-propose cycles), trust the confirmed set.
+                # re-propose cycles), trust the confirmed set. This path is
+                # safe for slice-scoped trackers because ``confirmed_roles``
+                # is the per-slice tracker's own state.
                 tracker_confirmed = tracker.confirmed_roles
                 if all_roles and all_roles.issubset(tracker_confirmed):
                     logger.warning(
@@ -691,6 +697,23 @@ class ConcurrentPhaseExecutor:
                     )
                     result["is_complete"] = True
                     result["fallback"] = "tracker_confirmed"
+                elif self._slice_id is not None:
+                    # Slice-scoped: the message-bus fallback below scans
+                    # ``store.get_messages(pipeline_id)`` pipeline-wide and
+                    # cannot distinguish slice-1's CONFIRMs from slice-2's,
+                    # so it would falsely declare consensus the moment a
+                    # fresh slice spawns roles whose names match an already-
+                    # confirmed prior slice (#2535). The in-memory per-slice
+                    # tracker is the authoritative source for slice work;
+                    # an empty fresh tracker correctly returns
+                    # is_complete=False here so the pipeline run loop keeps
+                    # polling.
+                    logger.info(
+                        "Skipping pipeline-wide message-bus fallback for slice-scoped tracker",
+                        pipeline_id=self.pipeline.id,
+                        slice_id=self._slice_id,
+                        blocking_agents=result.get("blocking_agents", []),
+                    )
                 else:
                     # Message-bus fallback: scan message store for
                     # CONSENSUS_CONFIRMED messages (#1471/#1615).
--- a/orchestrator/gateway_client.py
+++ b/orchestrator/gateway_client.py
@@ -10,6 +10,7 @@ Provides coordination between the orchestrator and gateway sidecar for:
 
 import json
 import os
+import re
 import socket
 import subprocess
 import sys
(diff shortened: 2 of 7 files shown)


jwbron:
Thanks for the careful review. Pushed 4726e829 to address it.

**Issue 1 (test syntax, blocking)** — replaced the unreachable try/except block with `PipelineConfig(concurrent_execution=True)`. You're right that `concurrent_execution` is a normal Pydantic bool field; the original block was both dead and used Python-2 catch-and-bind syntax that 3.14 happens to parse as a tuple expression. Gone.

**Issue 2 (slice-unaware `_existing_confirmed_for_role`, blocking)** — went with option A (filter on `metadata["slice_id"]`) because option B (noop the idempotency check when `slice_id is not None`) would let a retry-looping slice agent spray duplicate CONFIRMs into the bus within its own slice, which is the failure mode #1890 was tightening. The change:

- `_existing_confirmed_for_role(pipeline_id, agent_role, phase, slice_id=None)` now filters by `metadata.get("slice_id") == slice_id` (None matches None, preserving the legacy non-slice path exactly).
- Both per-slice CONFIRMED writers (final at signals.py:1702 and `pending_acks` at signals.py:1681) now tag `metadata["slice_id"]` when `slice_id is not None`.
- Caller at signals.py:1672 passes `slice_id=slice_id`.
- Three new regressions in `test_consensus_confirmed_idempotent.py` cover (a) slice-2 first CONFIRMED not deduped by slice-1, (b) within-slice idempotency still works, (c) pipeline-scoped callers ignore slice-scoped CONFIRMs.
- Inline docstring on `_existing_confirmed_for_role` references #2409 the way the existing reconstruction comment does.

**Issue 3 (other slice-unaware paths, non-blocking)** — flagged in the PR body under "Out of scope (tracked separately)" with the specific call sites (`kubernetes_monitor.py:915,934`, `startup_reconciliation.py:308,312`, `routes/pipelines.py:4166,4183`, the two tier-1 health checks). Pointed at #2409 since its fix sketch already covers `slice_id` on `Message` + a slice-aware `reconstruct_tracker_from_messages` — the same lever that closes those sites in one sweep. Did not file a sibling issue (skipped to keep this round narrowly review-driven; happy to file one if you'd prefer separate tracking).

**Issue 4 (`_build_rebase_onto_args` docstring drift)** — softened "Mirrors" to "Performs the same ref-shape sanity checks as", and called out the two real differences (stripped argv, no `validate_git_args` call). The audit boundary is the gateway server, not the helper.

**Issue 5 (stale docstring in `stacked_pr_reconciler.py`)** — updated to point at `orchestrator.gateway_client._build_rebase_onto_args` with a one-sentence note explaining why it's inlined and why the security floor is unchanged.

**Issue 6 (404 retry storm risk, non-blocking)** — not addressed in this PR. The slice-tracker-missing 404 branch is correct (false-confirming via the bus is worse), and verifying the agent-side retry backoff is its own piece of work.

**Test coverage gap #1 (module-load `gateway*` import paranoia)** — not addressed; would need a pytest config tweak rather than a per-test patch. Worth a follow-up if we want stronger coverage but didn't feel review-blocking.

**Test coverage gap #2 (positive slice-N tracker test)** — covered by the three new tests in `test_consensus_confirmed_idempotent.py` (the first asserts slice-2's CONFIRMED is actually written, with `slice_id` metadata, and that the marker is created).

Issue #2039, opened 2026-04-24
Gateway filtered_push silently strips agent-restricted file edits to deletions instead of rejecting or reverting


jwbron:
## Symptom

When an agent commit modifies a file that lives under an agent-restricted path (e.g., \`.github/\`, \`Makefile\` in some role configs, etc.), the gateway's filtered-push pipeline does NOT reject the push and does NOT revert the file to its baseline content. Instead, it removes the file entirely from the resulting tree — leaving a commit on the shared pipeline branch whose effect is \"delete the restricted file\".

In pipeline \`issue-1973\` implement phase, the coder agent legitimately tried to modify \`.github/workflows/test.yml\` per plan TASK-4-1 (change \`make test PYTEST_ARGS=...\` → \`make test-all PYTEST_ARGS=...\`). The push was accepted; the auto-filter stripped the entire file. The resulting branch commit \`c2897efb6\` carries an \`Auto-Filtered: true\` trailer and a single hunk \`-@@ -1,79 +0,0 @@\` — every line of \`test.yml\` deleted, with no replacement.

If reviewer_contract had not caught this on careful diff review, the resulting PR to \`main\` would have **removed the unit-test CI workflow entirely** (no pytest on PRs, no coverage gate).

## Reproducer

In any pipeline that has a coder/tester/etc. role with \`.github/\` in its restriction patterns:

\`\`\`
$ # Inside the agent sandbox:
$ # Edit .github/workflows/test.yml (legitimately, per plan)
$ git add .github/workflows/test.yml
$ git commit -m \"implement(#X): switch CI to make test-all\"
$ git push origin HEAD:egg/issue-X
$ # Push succeeds. Filter rewrites the commit.
$
$ # On the host:
$ git fetch origin egg/issue-X
$ git show <new-tip> -- .github/workflows/test.yml
$ # Shows full deletion, not the intended single-line change.
$ git show <new-tip> | grep Auto-Filtered
$ # \"Auto-Filtered: true\"
\`\`\`

## Root cause

\`gateway/filtered_push.py:_filter_tree\` (lines 207–208 as of \`origin/main\` HEAD ~ 2026-04-24):

\`\`\`python
# When a blocked path was modified, remove it from the tree:
subprocess.run([\"git\", \"update-index\", \"--force-remove\", blocked_path], ...)
\`\`\`

[...]

PR #2931, opened 2026-06-02
fix(gateway): exempt infra-authored commits from restricted-path block (#2927)

### LINKED ISSUE #2927: Gateway restricted_path_modified blocks a producer's proposal on commits by unregistered/bot authors (mis-attributed as own-authored); producer can't self-recover (#2039)
## Summary

On a sliced / event-pump implement phase, a **producer that completed its tasks cannot land `mcp__brc__propose`** because the gateway's role-based path-ownership check ([#2039](https://github.com/jwbron/egg/issues/2039)) rejects the push with `403 restricted_path_modified` for files the producer never authored.

The blocked files' *latest-touching* commits were made by **unregistered git identities** (the `egg-orchestrator` bot and the `egg-localhost` auto-formatter) that have **no entry in the commit-authorship registry**. Attribution therefore can't classify them as "pulled-from-other-role", so they fall into the **own-authored** bucket and the gateway blocks the push. The producer cannot self-recover, burns the cycle, files an impasse, exits, and the consensus-wrapper restart-loops it.

## Observed (issue-2908-impl2, slice-3, event-pump enabled, 2026-06-01)

- coder completed task-3-1/3-2/3-3 (commits `912e5bb07`, `68913fa69`); local tests green (114/114, 147/147, 11/11, 31/31). Its own commits touch only source files (`orchestrator/consensus_wrapper.py`, `orchestrator/routes/pipelines.py`, `sandbox/egg_agent_tools/handlers/event_prompt.py`).
- `mcp__brc__propose` → `403 restricted_path_modified`, **5 blocked_paths**, none authored by the coder:
  - `.egg-state/brc-history/issue-2908-impl2-implement-slice-2.json`
  - `.egg-state/brc-history/issue-2908-impl2-implement-slice-2.md`
  - `orchestrator/tests/test_consensus_next_action_adversarial.py`
  - `tests/sandbox/egg_agent_tools/test_handlers_brc_memory_adversarial.py`
  - `tests/sandbox/egg_lib/test_orch_cli_brc_adversarial.py`
  - payload had `attribution_fallback: false` and a `pulled_commits` list with tester/documenter authorship.
- The latest-touching commit of the blocked test files (`066b9d555`) has git-author `egg` and **no commit-authorship registry entry**.
- Recovery attempts all failed: `rebase --onto` to drop the offending commits → merge conflicts at `05a1a110e` **and rewrote slice-2's registry-recorded SHAs, escalating the blocked set 5 → 11**; `git branch -f`, `git checkout -B`, `git update-ref` → all blocked by pipeline-session guardrails ("Branch switching is not allowed in pipeline sessions").
- coder filed `impasse(category=external_blocker)` and exited; consensus-wrapper logged `Agent exited without BRC consensus. Restarting (1/3)...` → **restart loop**.

## Why #2039's design doesn't hold here

The gateway README (#2039) states: attribute each commit via the authorship registry, partition own-authored vs pulled-from-other-role, check write perms against **only** the own-authored set, and "**pulled cross-role commits never block the push**." That invariant holds only when every commit is attributable. When a blocked path's latest-touching commit is by an **unregistered author** (orchestrator bot / auto-formatter) — or by a commit whose SHA was rewritten by a recovery rebase and thus dropped out of the SHA-keyed registry — attribution can't classify it as pulled, it defaults into the own-authored bucket, and it blocks. So **infra/bot/auto-formatter commits on a shared slice branch can wedge any producer's proposal**, and the prescribed manual recovery is unreachable inside a locked pipeline session.

## Impact

Producers wedge on `propose` even with correct, complete work; tens of minutes burned; the consensus-wrapper restart-loops the agent (1/3, 2/3, …); the slice cannot advance without operator intervention.

## Suggested directions (non-prescriptive)

Note #2039 deliberately replaced the destructive silent-strip auto-filter from [#1882](https://github.com/jwbron/egg/issues/1882), so prefer **non-destructive auto-resolution** over re-introducing silent strips:

- Treat commits by known infra/bot identities (`egg-orchestrator`, `egg-localhost` auto-formatter) and registry-absent authors as **not own-authored** by the pushing role — exempt them from the restricted-path block.
- Consider **patch-id / content-based attribution** so rebase-rewritten pulled commits retain their original `author_role` (today, rebasing to recover breaks SHA-keyed attribution and *escalates* the block).
- When the own-authored set is empty but registry-absent commits touch restricted paths, the gateway should **auto-resolve** (allow the push, or auto-attach a conditional-ACK pre-merge condition) rather than `403` + an unreachable manual recovery.
- The documented recovery ("drop the edits and re-propose") is **not executable** inside a locked pipeline session for inherited commits — either make it executable or stop recommending it.
- Operator workaround the agent itself identified: forward-merge the parent slice branch into the child preserving registered SHAs so the child's diff contains only the producer's own files.

## Refs

- `gateway/gateway.py:1601-1764` (restricted-path partitioning / enforcement, `restricted_path_modified`)
- `gateway/README.md` — Agent-role restrictions section (#2039); conditional-ACK pattern #1998; prior auto-filter #1882
- Surfaced while operating `issue-2908-impl2` (slice-3 coder).

jwbron:
## Summary

Fixes #2927. A producer on a sliced/event-pump implement phase could not land `mcp__brc__propose`: the gateway's #2039 role-based path-ownership check rejected the push with `403 restricted_path_modified` for files the producer **never authored**.

### Root cause

The gateway attributes each pushed commit via the SHA-keyed commit-authorship registry, partitions files into own-authored vs pulled-from-other-role, and checks write perms against only the own-authored set ("pulled cross-role commits never block").

Infra commits — the `egg-orchestrator` state-file committer (`orchestrator/entrypoint.sh`) and the auto-formatter — are created **outside any agent session**, so the inline commit observer (`commit_observer.py`) never registers them. Attribution returns `None` for their SHAs, and the push handler buckets `None` as **own-authored (fail-closed)** (`gateway.py:1691-1695`, `git_client.py`). So a producer whose shared slice branch merely *inherited* those commits wedges: blocked for files it didn't touch, with every documented recovery unreachable inside the locked pipeline session (branch-switch denied; rebase rewrites SHAs and *escalates* the block).

### Fix

In `get_attributed_changed_files_in_push`, for SHAs the registry returns nothing for, read the **committer email** and, if it matches the trusted infra allowlist (`egg@localhost`, `egg@example.com`, `egg-salvage@localhost`), tag the commit with a reserved `infra` role. Those commits then land in the pulled partition and never block the push.

### Why this is safe (keeps #2039's fail-closed intent)

- Every **agent** commit uses `{role}@egg.local` (set whenever `EGG_AGENT_ROLE` is present, which the orchestrator always injects), so no agent identity collides with the infra allowlist.
- Committers **outside** the allowlist stay unregistered → own-authored → fail-closed. Unknown authors are never exempted.
- The allowlist is the **safe failure direction**: an operator overriding `EGG_USER_GIT_EMAIL` only *loses* the exemption (push fails closed as before), never gains a bypass.
- The commit observer is bypass-proof and registers every in-session commit to its role, so a forged infra committer is neutralized — attribution returns the agent's own role and still blocks. The only residual window (forged infra committer **and** concurrent registration failure) already fails closed for the agent's own `@egg.local` files.

### Scope / follow-up

Out of scope (separate, larger change): the **rebase-rewrite escalation** — recovery rebases mint new SHAs that drop out of the SHA-keyed registry. That needs patch-id/content-based attribution and is tracked as a follow-up.

## Tests

- `gateway/tests/test_git_client_attribution.py` — new `TestInfraIdentityExemption`: each infra email (incl. case-insensitive) → reserved `infra` role; unknown committer stays `None` (fail-closed); registered commit skips the committer lookup; `git show` failure stays `None`; mixed agent+infra range attributes each correctly.
- `gateway/tests/test_push_author_attribution.py` — new scenario 7b: an `infra`-attributed commit is treated as pulled (plain push, surfaced in `pulled_commits`), not blocked.
- Full `make test` suite green (17098 passed, 34 skipped); `make lint` clean.


jwbron:
Follow-up for the rebase-rewrite escalation case (patch-id/content-based attribution): #2932

jwbron:
Deferred follow-up tracked in #2932 (rebase-rewrite escalation → patch-id/content-based attribution).

PR #2998, opened 2026-06-04
fix(deploy): bound docker store before k3s-import so GC stops evicting fresh images

jwbron:
## Problem

`make redeploy` intermittently fails `check-egg-images-present` reporting the freshly-built `egg-*:<tag>` images are *not in k3s* — **even when HEAD hasn't moved** (the tag is current). Re-running the recommended `sudo k3s crictl rmi --prune` and redeploying fixes it briefly, then it recurs.

## Root cause — the wrong store was being drained

egg keeps **two** image stores on **one shared root fs**:

| Store | Typical size |
|---|---|
| k3s containerd (`/var/lib/rancher/k3s/agent/containerd`) | ~4 GB |
| docker (`/var/lib/docker`) | **tens of GB** |

`reap-stale-egg-images.sh` bounds the **first** after every deploy. Nothing bounded the **second**. So docker accumulates a full egg image set (including the ~12 GB `egg-sandbox`) for every `git describe` tag ever built, plus an unbounded BuildKit cache — keeping the root fs parked near kubelet's `imageGCHighThresholdPercent` (~85%).

`k3s-import` then does, per image, `docker save > /var/tmp/*.tar` **+** `k3s ctr images import` — briefly a *second and third* copy of the sandbox image on the same root fs. That transient spike crosses 85%, kubelet image GC evicts the freshly-imported, **not-yet-referenced** egg images mid-import (the running pods are still on the old tag), and `check-egg-images-present.sh` aborts the deploy.

The previously-recommended `k3s crictl rmi --prune` only reclaims the ~4 GB k3s store, so it never moved the disk needle — hence the recurrence. Diagnostic on the affected host: root fs 82%, docker 77 GB with ~33 GB reclaimable vs. k3s containerd 4.3 GB; only `egg-litellm` (imported *last*, after the spike cleared) survived in containerd while gateway/orchestrator/sandbox were evicted — the signature of GC eviction, not import failure.

## Fix

Add `scripts/reclaim-docker-disk.sh` and call it at the top of `k3s-import`, keyed on the tag about to be imported:

- reap stale `egg-*:<tag>` docker images (keep the import tag + `:latest`)
- prune dangling images
- cap the BuildKit cache via `--max-used-space` (default 20 GB, override `EGG_DOCKER_CACHE_MAX`)

This frees the prior tag's ~12 GB and excess cache **before** the import spike, so it stays under the GC line, and runs on every redeploy — bounding docker the way the containerd reap bounds k3s.

### Notes
- **Tag-scoped removal, no digest guard (by design).** `docker image rm repo:tag` untags rather than deleting a shared image ID, so — unlike the containerd reap, which removes by image ID and *needs* its digest guard — this one is inherently safe without one. Documented in both scripts so neither gets "unified" into the other.
- **Four-image safety gate.** Skips the reap unless every `egg-*:<tag>` image is present, so a half-built tag never strands the import with nothing to save.
- **Best-effort.** A reclaim hiccup never fails the import (`|| true`, plus per-removal tolerance). Needs no sudo — the docker socket is group-accessible.

## Testing
- `tests/scripts/test_reclaim_docker_disk.py` — 6 tests via a PATH-shimmed `docker`: safety gate, stale-tag reap, keep-tag/`:latest` protection, rm-failure tolerance, cache cap, no-docker no-op. All green.
- `ruff`, `shellcheck --severity=warning`, and `shfmt -i 2 -ci -bn` clean on the new files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -676,6 +676,18 @@ sudo-keepalive:
 # image is ~12 GB, so /tmp (tmpfs) would consume that much RAM.
 k3s-import: SHELL := /bin/bash
 k3s-import: sudo-keepalive  ## Import built images into k3s
+	@# Reclaim docker disk BEFORE the import. The loop below does, per image,
+	@# `docker save > /var/tmp/*.tar` + `ctr images import` -- briefly a second
+	@# AND third copy of the ~12 GB sandbox image on the shared root fs. If
+	@# docker's accumulated old tags + unbounded BuildKit cache already sit near
+	@# kubelet's image-GC high-water mark (~85%), that spike trips GC, which
+	@# evicts the freshly-imported, not-yet-referenced egg images mid-run and
+	@# aborts the deploy (see check-egg-images-present.sh). Bounding docker here,
+	@# keyed on the tag we are about to import, keeps the spike under the line.
+	@# Best-effort: a reclaim hiccup must not fail the import. Complements the
+	@# post-deploy reap-stale-egg-images.sh, which bounds the OTHER store (k3s
+	@# containerd); this one bounds docker.
+	@scripts/reclaim-docker-disk.sh "$(EGG_IMAGE_TAG)" || true
 	@set -euo pipefail; \
 	tmp=$$(mktemp -d -p /var/tmp egg-k3s-import.XXXXXX); \
 	trap 'rm -rf "$$tmp"' EXIT; \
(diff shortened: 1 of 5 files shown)


jwbron:
Closing in favor of #2999.

Investigation past this PR's premise found the recurring `make redeploy` failure is **btrfs chunk over-allocation** from large per-redeploy image churn — not the docker store crossing the 85% image-GC line, which was a symptom-level read. kubelet GC deleting images doesn't return btrfs chunks to *unallocated*, so DiskPressure sticks (`freed 0 bytes`) regardless of docker-store size.

The docker-reclaim work here (branch `egg/bake-docker-reclaim-into-make`) is still useful as lever **C** in #2999 (bound the docker store to slow accumulation), but it's hygiene, not the fix. The real lever is **A** — skip the 12 GB sandbox save+import when its content is unchanged — plus **B**, a btrfs `balance` reclaim step. Branch left intact for revival under the corrected framing.

Issue #2999, opened 2026-06-04
make redeploy recurringly wedges local k3s (DiskPressure / 'images not in k3s') — root cause: btrfs chunk over-allocation from image churn


jwbron:
## Summary

On the local single-node k3s dev host (Fedora Asahi, `/` on **btrfs**), `make redeploy` recurringly fails at `check-egg-images-present` ("egg-system images for tag X are not in k3s") and/or wedges the node with `DiskPressure=True` — all pods evicted, gateway/orchestrator/litellm deployments `0/1`, nothing deploys or runs until it's cleared by hand.

Extended investigation traced this **past** the surface symptom (kubelet image-GC evicting freshly-imported images) to the real disease: **btrfs data-chunk over-allocation** driven by the large per-redeploy image churn.

## Root cause

`make k3s-import` re-`docker save`s + `ctr import`s the full egg image set every redeploy — including the **~12 GB `egg-sandbox`**, which is briefly *triple-resident* on the shared root fs (docker layer store + `/var/tmp` tarball + containerd). That create/delete churn fragments btrfs data chunks. btrfs ends up with **every chunk allocated but most of them near-empty** (observed: device `129.57/129.57 GiB` allocated, **~1 MiB unallocated, only ~68 GiB actually used**).

Because `statfs` (what kubelet's imagefs accounting reads) counts allocated-but-empty chunks as used, the node reports ~86% usage → kubelet image-GC fires (`> 85%` high threshold) → evicts the freshly-imported, not-yet-referenced egg images → `check-egg-images-present` then reports them missing. And deleting images **does not return chunks to unallocated**, so GC reports `freed 0 bytes` / "failed multiple times in a row," DiskPressure sticks, and pods get evicted.

## Evidence

kubelet (k3s) logs during a failed redeploy:
```
image_gc_manager.go:528] "Removing image to free bytes" imageID="sha256:a99f…089d64" size=12451223127   # 12.45 GB = egg-sandbox, GC'd mid-import
image_gc_manager.go:394] "Disk usage on image filesystem is over the high threshold, trying to free bytes
                          down to the low threshold" usage=86 highThreshold=85 amountToFree=6973085286 lowThreshold=80
kubelet.go:1652] E "Image garbage collection failed multiple times in a row"
                   err="Insufficient free disk space … (86% of 129.6 GiB used). … freed 0 bytes"
```
btrfs: `Device allocated 129.57/129.57 GiB, unallocated 1 MiB` while only ~68 GiB used.

[...]

PR #3109, opened 2026-06-11
feat(deploy): include egg-sandbox in the registry publish path by default

jwbron:
Follow-up to #3101 (refs #2999).

#3101 shipped the loopback-registry publish flow with the sandbox image excluded from `EGG_REGISTRY_IMAGES` pending verification that its private repo content could never be published off-host. That's now confirmed and accepted: the registry is a `registry:2` container on the dev host **bound to 127.0.0.1 only** (`setup-local-registry.sh` refuses to set up anything else), and `push-egg-images.sh` **hard-refuses any non-loopback registry with no override** — exposure is identical to the docker daemon store and k3s's containerd.

With that settled, this flips the default so `egg-sandbox` rides the registry path too and gets the same layer-incremental publish as the core images. The practical effect is the rest of the #2999 win: a code-only redeploy moves the changed ~25 MB of layers instead of a full ~5.3 GB sandbox `docker save` + `ctr import` per edit.

- One functional line: `EGG_REGISTRY_IMAGES ?= $(EGG_ALL_IMAGES)`; everything else is comments/docs catching up.
- The exclusion knob remains: remove an image from `EGG_REGISTRY_IMAGES` and it publishes via save+import (store-to-store, no registry). `EGG_IMAGE_REGISTRY=` (empty) still disables the registry flow entirely (CI is pinned there).
- Verified: `make -n` renders in default (all four pushed, import leg empty-guarded), exclusion, and no-registry modes; shellcheck clean; no behavior change for CI.


jwbron:
Pushed two fixes for the failure seen on the first live redeploy (`short read: expected 2552 bytes but got 0` on the orchestrator pre-pull):

1. **Root cause**: registry:2 serves blob-existence checks from an in-memory descriptor cache, while `garbage-collect` (run via `docker exec` by the post-deploy reap) deletes blob files behind the serving process's back. The next push gets "Layer already exists" for a blob whose file is gone, skips the upload, and the subsequent containerd pull hits a truncated read. The reap now does a ~1s `docker restart egg-registry` right after GC so existence checks hit disk honestly.
2. `k3s-publish` now propagates a `k3s-push` failure (`|| exit 1`) instead of letting the import leg's exit code mask it — previously a failed push could still let `deploy` repoint the cluster at unpullable images.

The host registry has been healed in place (restart + re-push of all core images; every blob of every tag verified servable at full size via the registry API). Note the pre-pull guard worked exactly as designed — it failed the publish *before* `kubectl apply` touched the cluster.

Issue #2804, opened 2026-05-27
Agent SDK message reader has 1MB JSON buffer limit — large tool results crash agents with exit code 255


jwbron:
## Summary

The egg-agent runtime / Anthropic agent SDK message reader has a
hard-coded **1MB (1,048,576 bytes) JSON message buffer limit**. When a
tool result (e.g. `Bash` output from a deep `grep`, `git diff` of a large
diff, `Read` of a long file, MCP tool result with a substantial payload)
exceeds this limit, the JSON reader fails with:

```
Fatal error in message reader: Failed to decode JSON: JSON message
exceeded maximum buffer size of 1048576 bytes...
```

The agent then exits with **code 255**. The consensus-wrapper attempts
**up to 2 restarts** (per `Restarting (1/2)...` / `(2/2)...`); if both
restarts encounter the same overflow (or any other crash), the agent
stays dead and the phase BRC consensus stalls.

When this happens to multiple concurrent agents in the same phase
(common in implement slices with 8 reviewers + 3 producers all running
`grep` / `find` / `Read` on the same codebase), the phase enters
terminal `failed` status with no producer proposals reaching consensus.
This presents as:

- All N containers exit with code 255 within seconds of each other.
- `recent_messages` shows `OVERSEER_ALERT overseer_restart: overseer [info]`
  followed by `AGENT_FAILED` for one or more roles.
- `OVERSEER_ALERT stuck-phase-transition [high]` fires shortly after.
- A HITL `decision-N` is raised with options
  `[\"Retry phase\", \"Accept current state\", \"Abort phase\"]`.

## Observed on `issue-2777`

The SDLC pipeline cleaning up the sliced implement phase has hit this
**three times** in a single run:

1. **Plan phase crash, 2026-05-27 ~06:02 UTC** — all 5 plan-phase
   agents (architect, task_planner, risk_analyst, reviewer_plan,
   overseer) exited code 255 simultaneously. No clear root cause in
   logs at the time, but matches this pattern. Recovered via
   `restart_phase implement`.

2. **First implement-slice-1 crash, 2026-05-27 ~11:45 UTC** — all 8
   slice-1 agents exited code 255. Pipeline went `failed`; `decision-14`
   was raised. Operator selected \"Retry phase\" (which separately
   doesn't auto-invoke restart — see #2801). After explicit
   `restart_phase`, the slice re-spawned.

3. **Second implement-slice-1 crash, 2026-05-27 16:12 UTC** — coder
[...]

Issue #1762, opened 2026-04-21
Replace interactive mode with a generic custom-phase MCP primitive


jwbron:
## Motivation

The `egg` CLI's default path is an **interactive mode** (`run_claude()` in `sandbox/egg_lib/runtime.py:634`) that spawns a Claude Code session inside a sandboxed container, after bringing up gateway+orchestrator via `docker-compose`. This predates the MCP server and no longer earns its complexity.

All egg agents should be headless. A *host session* (Claude Code on the user's machine, or any MCP client) should trigger one-off agent work through the MCP server and consume the result — rather than users dropping into a sandboxed interactive Claude session.

The replacement primitive is a **custom phase**: a single phase run against a repo, with an explicitly specified subset of that phase's agent roles. BRC consensus always applies; degenerate rosters (one producer, no reviewers) short-circuit on the first proposal.

## What to remove

- `run_claude()` and its supporting code in `sandbox/egg_lib/runtime.py`
- The interactive default branch in `sandbox/egg_lib/cli.py:215` and flags that exist solely for interactive use (`--setup`, `--reset`, `--public`/`--private` session modes, `--compose --down/--build`)
- The Docker Compose base workflow: `sandbox/egg_lib/compose.py` and the compose paths in `bin/egg-deploy`. With k8s as the runtime for gateway+orchestrator, compose is no longer the deployment model.
- The interactive branch of `sandbox/entrypoint.py` (`run_interactive()` at line 1907, dispatch at line 2141) — entrypoint becomes exec-only.
- The `bin/egg` top-level binary entirely. It's a thin wrapper over `egg_lib.cli.main` whose only purpose is launching interactive mode. The other `bin/egg-*` scripts (`egg-deploy`, `egg-sdlc`, `egg-status`, `egg-pipeline-watch`, `egg-onboarding-docs`) are unaffected.

### Shape

A new pipeline mode `PipelineMode.CUSTOM` (alongside `ISSUE` and `BABYSIT` in `orchestrator/models.py:29`) and a new MCP tool `run_custom_phase` in `orchestrator/mcp_tools.py`.

```
run_custom_phase(
  phase: "refine" | "plan" | "implement",    # required
  roles: list[str] | None,                   # subset of phase's roles; default = full phase roster
  repo: str,                                 # required
  description: str,                          # free-form task description
  branch: str | None,                        # target branch (new if absent)
  base_branch: str | None,                   # defaults to repo default branch
  pr_number: int | None,                     # optional existing PR target
[...]

PR #3137, opened 2026-06-11
fix(sandbox): keep Claude Code current so built-in model aliases resolve (#3136)

### LINKED ISSUE #3136: Sandbox Claude Code rejects the 'fable' model alias — refine/plan agents crash-loop at spawn on the built-in default
## Symptom

Pipeline `issue-3077` (2026-06-11): the refiner crash-looped from spawn — **160+ consecutive failed invocations**, one every ~2–3 s, each dying instantly with:

> There's an issue with the selected model (fable). It may not exist or you may not have access to it. Run --model to pick a different model.

Per-invocation log signature (`egg-agent: Agent completed ... model=<synthetic> ... success=False`), followed by `[event-pump] agent invocation failed (action=propose, rc=1, streak=N). Idle counter continues to accrue.` The phase made zero progress; the only detection was the overseer's heartbeat-silence alert ~8 minutes in.

## Root cause

`orchestrator/agent_model_resolution.py` defaults the refine/plan-phase roles (`_FABLE_DEFAULT_ROLES`: architect, refiner, reviewer_agent_design, reviewer_plan, reviewer_refine, risk_analyst, task_planner) to `FABLE_DEFAULT_MODEL = "fable"` at tier 3. The sandbox's installed Claude Code build **does not recognize the bare `fable` alias**.

Isolation evidence — same sandbox image, same mounted auth:

- `model=fable` → instant "may not exist or you may not have access" on every invocation.
- `model=claude-fable-5` (versioned ID) → session initializes and the agent runs normally (verified on the resubmitted `issue-3077` pipeline).
- The **host** Claude Code 2.1.173 resolves `"model": "fable"` from settings.json without issue.

Since the exact versioned ID works under the same credentials, this is an **alias-table gap in the sandbox's Claude Code build**, not an API entitlement problem. The sandbox Dockerfile installs CC via `ARG CLAUDE_CODE_VERSION=stable` at image build time (sandbox/Dockerfile:249-250), so the deployed image's CC predates the `fable` alias. (Hypothesis from the install mechanism; the image's `claude --version` wasn't directly inspected.)

## Workaround used

Resubmitted with per-role overrides:

```json
{"agent_models": {"architect": "claude-fable-5", "refiner": "claude-fable-5", "reviewer_agent_design": "claude-fable-5", "reviewer_plan": "claude-fable-5", "reviewer_refine": "claude-fable-5", "risk_analyst": "claude-fable-5", "task_planner": "claude-fable-5"}}
```

Known side effect (documented in agent_model_resolution.py:71-79): the `FABLE_EFFORT = "high"` pin keys on the bare `fable`/`fable[1m]` aliases only, so `claude-fable-5` routes through the generic `claude-*` path with `effort=None` — confirmed in the session-init log (`model=claude-fable-5 effort=`). The versioned name escapes the effort-drift defense the alias pin exists to provide.

## Suggested fix

1. **Refresh the sandbox image's Claude Code** to a build whose alias table includes `fable`, and verify at build time (e.g. fail the image build if `claude` can't resolve the aliases that `agent_model_resolution.py` uses as built-in defaults — `fable` and `opus` — rather than discovering it at spawn time in a live pipeline). Keeping the default on the bare alias is right per the no-version-pinning policy; the bug is the stale runtime, not the alias.
2. Until the image is refreshed, the tier-3 default is a guaranteed crash-loop for every refine/plan phase on the current deploy — worth a loud operator-facing note or a preflight check.

## Secondary observation

The event pump retried a deterministic, permanent failure (unknown model — fails in <1 s, rc=1) **160+ times with no backoff and no failure classification**; the streak counter exists but nothing acts on it for the `propose` arm, and detection fell to the overseer's generic heartbeat-silence anomaly ~8 minutes in. A fast-failing invocation streak in the hundreds is a strong fail-fast/backoff signal independent of this particular model bug.

## Incident reference

First `issue-3077` pipeline run, 2026-06-11 ~20:31–20:45 UTC (cancelled, cleanup; stale `egg/issue-3077/work` branch had to be deleted manually before resubmission). Resubmitted run with the `claude-fable-5` overrides is the live `issue-3077` pipeline.

jwbron:
Fixes #3136.

## Problem

The first `issue-3077` run crash-looped every refine/plan agent at spawn (160+ failed invocations): the sandbox image's Claude Code build doesn't know the bare `fable` alias that `orchestrator/agent_model_resolution.py` uses as the tier-3 default for refine/plan roles.

Root mechanism: `make build` runs plain `docker build`, so `ARG CLAUDE_CODE_VERSION=stable` is the operative value (the legacy egg CLI that passed a concrete version was removed in #1762). The install layer is keyed on the literal string `stable`, BuildKit never re-runs it, and the image's CC froze at whatever stable pointed to when the layer first built.

Compounding finding: the **stable channel itself doesn't have `fable` yet** — current stable is 2.1.153, whose binary contains zero `claude-fable` strings (verified by downloading and grepping the linux-arm64 release). `latest` (2.1.173, what the host runs) resolves it fine. So a rebuild against stable today would still crash-loop.

## Changes

**`sandbox/Dockerfile`** — three changes to the CC install:

1. **Default channel `stable` → `latest`.** egg's built-in defaults track new model families at launch; the stable channel lags those launches by weeks.
2. **Cache-bust via remote `ADD` of the channel manifest** (`downloads.claude.ai/claude-code-releases/latest` — the same manifest `install.sh` reads). BuildKit re-checks the URL content on every build, so the install layer re-runs exactly when the channel moves to a new version, and stays cached otherwise.
3. **Build-time alias gate** (issue's suggested fix 1): fail the image build if the installed binary doesn't know the model families used as built-in defaults (`fable`, `opus`), rather than discovering it at spawn time in a live pipeline. No auth is mounted at build time, so the gate is a strings-level heuristic: a build that knows a family alias embeds the versioned `claude-<family>-*` IDs the alias maps to (48 occurrences in 2.1.173; 0 in 2.1.153).

**`claude-agent-sdk` 0.1.65 → 0.2.97** (`sandbox/pyproject.toml` pin `>=0.2.97,<0.3` + Dockerfile ARG). Per the pin's "validate manually on 0.2" note, I verified against 0.2.97 in a clean venv:

- all imported symbols present: `query`, `ClaudeAgentOptions`, `create_sdk_mcp_server`, `tool`, message/block types, permission results, hook types
- every `ClaudeAgentOptions` field egg sets exists (`permission_mode`, `setting_sources`, `can_use_tool`, `max_buffer_size`, `effort`, `mcp_servers`, `hooks`, …)
- `CLIJSONDecodeError` is still a `ClaudeSDKError` subclass and the `exceeded maximum buffer size` marker (#2804) is unchanged in the SDK source

`tests/sandbox/egg_agent_tools/test_sdk_surface.py` passes (pin-format test; the import smoke tests run in-sandbox where the SDK is installed).

## Not in this PR

The event pump retried the deterministic rc=1 failure 160+ times with no backoff or failure classification — filed separately as a follow-up (see issue link in comments).

## Testing

- `make lint` clean (hadolint on the modified Dockerfile included)
- SDK 0.2.97 surface validation as above
- Channel/alias facts verified directly against `downloads.claude.ai` release binaries

Note for deploy: the new CC lands on the next `make redeploy` (which rebuilds all images including egg-sandbox and publishes via the loopback registry, #2999/#3109), and the gate makes that rebuild fail loudly if the channel ever regresses.

### DIFF
--- a/docs/guides/per-agent-models.md
+++ b/docs/guides/per-agent-models.md
@@ -109,12 +109,13 @@ in `orchestrator/agent_model_resolution.py` walks the chain:
 > **Minimum sandbox Claude Code version.** The `fable` / `fable[1m]`
 > aliases require Claude Code ≥ 2.1.170. The sandbox's
 > `CLAUDE_CODE_VERSION` build-arg in `sandbox/Dockerfile` defaults to
-> `stable`, which satisfies this on a fresh build. Deployments that
-> pin an older `CLAUDE_CODE_VERSION` will see refine/plan agent spawns
-> fail with an "unknown model" error from Claude Code — either bump
-> the pinned version or set a repo-level `default_agent_model: opus`
-> to opt back out of the fable default until the sandbox image is
-> rebuilt.
+> `latest` (#3137 — `stable` lagged the fable launch and crash-looped
+> refine/plan agents per #3136). A build-time gate in the Dockerfile
+> fails the image build if the installed binary doesn't know `fable`
+> or `opus`, so a stale pin surfaces at build time rather than at
+> spawn. If you need to pin to an older `CLAUDE_CODE_VERSION` that
+> predates the alias, also set a repo-level `default_agent_model: opus`
+> AND temporarily drop `fable` from the Dockerfile gate.
 
 The result is an `AgentModelDecision` dataclass with fields
 `(claude_code_alias: str, upstream: str, upstream_model: str | None,
--- a/docs/reference/agent-tools.md
+++ b/docs/reference/agent-tools.md
@@ -545,7 +545,7 @@ complete shell CLI surface.
 
 ## Version pin
 
-`claude-agent-sdk` is pinned to `>=0.1.65,<0.2` in
+`claude-agent-sdk` is pinned to `>=0.2.97,<0.3` in
 `sandbox/pyproject.toml` and the `CLAUDE_AGENT_SDK_VERSION` ARG in
 `sandbox/Dockerfile`. A smoke test at
 `tests/sandbox/egg_agent_tools/test_sdk_surface.py` imports
--- a/sandbox/Dockerfile
+++ b/sandbox/Dockerfile
@@ -230,12 +230,14 @@ RUN groupadd -g 1000 egg && \
     mkdir -p /home/egg/.config && chown egg:egg /home/egg/.config
 
 # Install claude-agent-sdk (changes more often than base deps, less often than CLI)
-# The egg script always passes the specific PyPI version to bust Docker's layer
-# cache when a new version is released. The "latest" fallback only applies if the
-# build-arg is omitted (e.g. manual docker build without the egg script).
-ARG CLAUDE_AGENT_SDK_VERSION=0.1.65
+# Nothing passes this build-arg anymore (the legacy egg CLI that resolved a
+# concrete PyPI version was removed in #1762), so the default below is the
+# operative pin — bumping it is also what busts Docker's layer cache so the
+# new release actually installs. Keep it in sync with the bounded range in
+# sandbox/pyproject.toml.
+ARG CLAUDE_AGENT_SDK_VERSION=0.2.97
 RUN if [ "$CLAUDE_AGENT_SDK_VERSION" = "latest" ]; then \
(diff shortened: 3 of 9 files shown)


jwbron:
Follow-up for the event-pump no-backoff/no-escalation observation filed as #3138.

Issue #3077, opened 2026-06-10
Design: coordination state is served, not replicated — retire git/prompt choreography for agent state exchange


jwbron:
## The pattern

Laid end to end, a long lineage of incidents is one bug recurring: **pipeline coordination state has one logical home but N physical replicas (orchestrator worktree, gateway worktree, per-agent worktrees, origin, per-role/slice branches), and the replicas are synchronized by ad-hoc git choreography and prompt prose.**

| Incident | Replica failure |
|---|---|
| #3068 / #3073 | Seeded drafts on the work branch, agent worktrees forked from base — state never reached agents |
| #3072 | Narrow-refspec mirrors: fetch "succeeds," tracking refs stay stale — worktrees created behind the real tip |
| #3076 | `read_peer_artifact` reads `.egg-state/brc-history/` in the agent worktree; the file is only written at phase **end**, into the **orchestrator's** worktree — structurally empty mid-phase |
| #3033 | Spawn prompt (the channel carrying task text) discarded by the event pump (`del prompt_text`, consensus_wrapper.py:748) |
| #3016 | Canonical artifact paths live in prompt prose agents never reliably see — write-path drift (refiner; repeated by the architect's `architect-plan.md` in #3076) |
| #2629 | `populate_contract` wrote the orchestrator worktree but didn't push — fresh spawns couldn't see it |
| #2626 / #2625 | Plan-sync deleted drafts from the integration branch |
| #2972 / #2488 | Post-phase `_sync_worktree_with_remote` hard-reset discarded committed HITL decisions |
| #3005 | Gateway emptyDir wiped sessions → startup cleanup deleted live worktrees |
| #2548 / #2755 | brc-history slice partitions vs work branch: add/add merge conflicts |
| #3117 | Contract task-record mutations written only to the live worktree copy; the work-branch replica is frozen at init time — every task row reads `pending` forever (contributed to the #3114 misdiagnosis) |

[...]

Issue #2640, opened 2026-05-12
Integration test coverage: message store & event bus (Redis)


jwbron:
Expand k3s integration test coverage for the inter-agent message store and event bus on Redis. Recently changed in #2621 / #2624 (`context_pr.skipped`/`context_pr.failed` event routing); changes were unit-tested but the end-to-end routing isn't exercised on k3s.

## Starting points

1. **`context_pr.{skipped,failed}` routing** — drive a pipeline through the conditions that fire these events; assert they land in both message store and event bus.
2. **Wait-status semantics for the new event types** — block a consumer on `wait-status` for `context_pr.*`; assert correct unblock.
3. **Event ordering under concurrent producers** — multiple agents producing events; assert message store ordering matches event-bus ordering for a single agent's stream.

## Gap audit

Identify additional event-bus invariants — TTL/expiry, replay-on-resume, event-bus reconnection during agent restart, malformed-payload rejection. Add to the test list before implementing.

## Constraint

Integration tests run in CI, not inside the SDLC pipeline — the agent sandbox can't bring up k3s. Plans must treat "verified by green required-check" as the validation path. See #2474 for the full constraint write-up.

Split from #2474.

Issue #3002, opened 2026-06-04
Deploy egg on GKE (minimal first deploy): Terraform-provisioned cluster + GKE overlay + registry pull path


jwbron:
## Summary

Stand up a **minimal first deployment of egg on GKE**: provision a new GKE
cluster with Terraform (the GCP project already exists), get the egg control
plane (orchestrator + gateway + litellm) running on it, and prove the path by
driving **one real pipeline (`submit_task` → … → PR) end-to-end** on the
cluster with network isolation intact.

This is deliberately scoped as the *smallest* path to "egg runs in GKE." HA,
autoscaling, Secret Manager / Workload Identity hardening, ingress/TLS, managed
Redis, and an observability stack are explicit non-goals here and become
follow-ups (see below).

## Context: this builds on an existing k8s foundation

The Docker→Kubernetes migration ([#1553]) already landed a **runtime-agnostic
base**, so this is *not* a from-scratch k8s effort — it's a cloud overlay + a
registry pull path + cluster provisioning on top of what exists:

- **Kustomize `base/`** (`k8s/base/`) is environment-agnostic: namespaces
  (`egg-system`, `egg-agents`), orchestrator/gateway/litellm Deployments +
  Services, `network-policies.yaml`, and least-privilege `rbac.yaml`.
- **`k8s/overlays/local/`** holds all the k3s/single-node-specific patches
  (hostPath, hostPort) — the overlay seam GKE plugs into.
- The orchestrator already abstracts the runtime (`kubernetes_client.py`,
  `kubernetes_spawner.py`, `kubernetes_monitor.py`) and detects it at runtime
  (`EGG_RUNTIME`, `_detect_k3s`/`_detect_cni` in `orchestrator/routes/deployment.py`).
- **GKE was always the intended follow-up.** The migration doc scoped target env
  to "local k3s only" (design decision #7) and lists the GKE surface under
  "Future Work" (`docs/architecture/kubernetes-migration.md`).

GKE is also a *natural* fit for the existing NetworkPolicies: GKE **Dataplane V2
is Cilium-based**, and the policies were authored against Cilium, so they should
largely port rather than be rewritten.

## What's local-only today (the actual work surface)

The base is portable; the local overlay and the build/deploy path are not. The
known k3s/single-node assumptions a GKE deploy has to replace:

1. **Image distribution.** Local dev uses `docker save` → `k3s ctr images import`
[...]

PR #3153, opened 2026-06-12
Deploy Redis in k3s and pin the orchestrator to the Redis message store (#2662)

### LINKED ISSUE #2662: k3s does not deploy Redis — RedisMessageStore has no end-to-end CI coverage
## Problem

`orchestrator/message_store.py` auto-selects the Redis Streams backend (`RedisMessageStore`) when Redis is reachable, falling back to an in-memory `MessageStore` otherwise:

```python
use_redis = os.environ.get("EGG_MESSAGE_STORE_BACKEND", "auto")  # default: auto
redis_host = os.environ.get("REDIS_HOST", "localhost")
# ... tries to ping Redis; falls back to MessageStore() on failure
```

The k3s manifests under `k8s/base/` do not deploy Redis (`grep -rn redis k8s/` returns no Deployment / StatefulSet / Service / image), and `k8s/base/orchestrator-deployment.yaml` does not pass `REDIS_HOST` or `EGG_MESSAGE_STORE_BACKEND` to the orchestrator container. Result: the production k3s deployment always runs the in-memory `MessageStore`, never `RedisMessageStore`.

## Impact

- The `RedisMessageStore` codepath (XADD / XREAD BLOCK / XRANGE / xinfo counter, since_id cache + paginated-scan fallback, transient `RedisError` handling) has **zero live end-to-end coverage** in CI today. Coverage is unit-tier only, via `fakeredis` (`orchestrator/tests/test_redis_message_store.py`, plus the new integration tests in #2640 / PR #2661 which also use `fakeredis`).
- Messages are not persisted across orchestrator restarts (`MessageStore` is in-process). Any documentation that implies Redis-backed durability does not match the deployed behavior.
- `clear()` semantics on the in-memory backend are condition-variable based; the Redis backend uses `DEL` on the stream. A regression in either path could go unnoticed in the deployed environment until a swap occurs.

## Options

1. **Deploy Redis in k3s and switch the orchestrator to `EGG_MESSAGE_STORE_BACKEND=redis`** (or `auto` once Redis is reachable). Adds a StatefulSet + Service to `k8s/base/` plus the env wiring on `orchestrator-deployment.yaml`. The `make test-integration` workflow already deploys via `make deploy`, so adding the Redis manifests gets coverage for free.
2. **Document the in-memory backend as the supported production choice** and drop `RedisMessageStore` / its tests / `fakeredis` from dev deps. Loses the cross-restart durability story.

Option 1 is the lower-risk path if the durability story matters (#1897 implies it does). Option 2 if it doesn't.

## Out of scope for this issue

Test coverage for the routing on top of the message store is already addressed in #2640 / PR #2661 — what's missing is the *backend choice* in production.

jwbron:
Fixes #2662.

## Problem

The k3s manifests deployed no Redis and passed no `REDIS_HOST` / `EGG_MESSAGE_STORE_BACKEND` to the orchestrator, so `auto` backend selection always fell back to the in-memory `MessageStore`:

- `RedisMessageStore` had **zero live end-to-end coverage** (unit tier only, via fakeredis).
- Messages did not survive orchestrator restarts — the #3076 mid-phase-restart loss risk, and the replay source for consensus-tracker reconstruction.
- Since #3077 slice-6 (#3144), every deployed boot logged the `MESSAGE_STORE_AUTO_FALLBACK_TO_MEMORY` marker and `/api/v1/health` reported a permanently degraded `message_store` component.

The gap had a **third layer** the issue didn't name: the `redis` client package was not installed in the orchestrator image at all (absent from `orchestrator/requirements.txt` and pyproject runtime deps — only `fakeredis` existed, in dev deps). Deploying Redis + env alone would have changed nothing: the `import redis` inside `_create_message_store()` fails before the backend is ever probed.

## Changes

**egg-redis image** (`config/redis/Dockerfile`): stock Redis pinned at `8.6.4-alpine`, built and published through the local loopback-registry supply chain like every other egg image — no public image references in the manifests. Wired into `EGG_ALL_IMAGES`, `make build`, the deploy tag rewrite, the local overlay `images` list, and `await-egg-deploy.sh`.

**Manifests** (`k8s/base/redis-deployment.yaml`, `redis-service.yaml`): single replica, `Recreate` (never deadlocks on the RWO volume), 1Gi PVC with `appendonly yes` so messages also survive Redis pod restarts, `maxmemory 256mb` + `noeviction` so memory pressure fails writes loudly instead of silently dropping stream entries, hardened securityContext matching the existing deployments, `redis-cli ping` probes. Agent pods cannot reach it — `egg-agents` is default-deny egress with no Redis allow rule.

**Orchestrator wiring**: `EGG_MESSAGE_STORE_BACKEND=redis` — explicit, not `auto`. The silent fallback is exactly what the slice-6 degraded flag exists to catch; in explicit mode a Redis outage raises, and since the store singleton is created lazily and a failed creation leaves it unset, the orchestrator naturally retries until Redis is reachable (no startup-ordering dependency). Plus `REDIS_HOST`/`REDIS_PORT`, and `redis>=5.0,<7.0` in `orchestrator/requirements.txt` + pyproject runtime deps.

**End-to-end coverage** (`integration_tests/test_message_store_backend.py`), running against the cluster `make test-integration` deploys:
1. the deployed Deployment pins `EGG_MESSAGE_STORE_BACKEND=redis` (makes the health check non-vacuous — explicit mode cannot silently fall back);
2. `/api/v1/health` carries no fallback marker;
3. a live XADD/XRANGE/DEL round-trip from inside the orchestrator pod through the production selection path in `_create_message_store()` — exercising the env wiring, the in-image client, Service DNS, and the live Redis, and asserting the selected backend class is `RedisMessageStore`.

**Docs**: `concurrent-execution.md`'s claim that production stores messages in Redis Streams is now true; `coordination-state.md`'s wipe-semantics section notes the accidental-loss row can no longer occur in-cluster.

## Out of scope

Message-routing test coverage on top of the store (#2640 / #2661). Selection-precedence semantics are unchanged (frozen by #3077 HITL Q3) — this PR only changes deployment config, deps, tests, and docs.

## Validation

- `make lint` clean; `kubectl kustomize k8s/overlays/local/` renders with the image substitution and env in place.
- `orchestrator/tests/test_redis_message_store.py` + `test_message_store.py`: 123 passed.
- New integration tests collect cleanly; they exercise the live cluster on the next `make redeploy` / `make test-integration`.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -143,10 +143,12 @@ jobs:
               && docker save egg-orchestrator:latest | sudo k3s ctr images import - \
               && docker save egg-sandbox:latest | sudo k3s ctr images import - \
               && docker save egg-litellm:latest | sudo k3s ctr images import - \
+              && docker save egg-redis:latest | sudo k3s ctr images import - \
               && docker save egg-gateway:"$EGG_IMAGE_TAG" | sudo k3s ctr images import - \
               && docker save egg-orchestrator:"$EGG_IMAGE_TAG" | sudo k3s ctr images import - \
               && docker save egg-sandbox:"$EGG_IMAGE_TAG" | sudo k3s ctr images import - \
-              && docker save egg-litellm:"$EGG_IMAGE_TAG" | sudo k3s ctr images import -; then
+              && docker save egg-litellm:"$EGG_IMAGE_TAG" | sudo k3s ctr images import - \
+              && docker save egg-redis:"$EGG_IMAGE_TAG" | sudo k3s ctr images import -; then
               echo "::endgroup::"
               echo "Image import succeeded on attempt ${attempt}"
               exit 0
@@ -175,7 +177,7 @@ jobs:
         run: |
           set -x
           docker images --format '{{.Repository}}:{{.Tag}}' \
-            | grep -E '^egg-(gateway|orchestrator|sandbox|litellm):' \
+            | grep -E '^egg-(gateway|orchestrator|sandbox|litellm|redis):' \
             | xargs -r docker rmi -f || true
           docker builder prune -af || true
           docker image prune -af || true
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ EGG_IMAGE_PREFIX := $(if $(EGG_IMAGE_REGISTRY),$(EGG_IMAGE_REGISTRY)/,)
 # registry entirely, remove it from EGG_REGISTRY_IMAGES; excluded images
 # publish through the save+import path instead (slower, but entirely
 # store-to-store on this host, no registry involved).
-EGG_ALL_IMAGES := egg-gateway egg-orchestrator egg-sandbox egg-litellm
+EGG_ALL_IMAGES := egg-gateway egg-orchestrator egg-sandbox egg-litellm egg-redis
 EGG_REGISTRY_IMAGES ?= $(EGG_ALL_IMAGES)
 # Images the registry path does NOT cover (imported via k3s-import instead).
 EGG_IMPORT_IMAGES := $(filter-out $(EGG_REGISTRY_IMAGES),$(EGG_ALL_IMAGES))
@@ -518,7 +518,7 @@ build: sync-venv-if-uv
 	@$(PYTHON) scripts/prepare-sandbox-build-context.py repo-deps
 	@echo "==> Building images with tag $(EGG_IMAGE_TAG) ($(BUILD_JOBS) parallel jobs)..."
 	@$(MAKE) --no-print-directory -j$(BUILD_JOBS) $(BUILD_OUTPUT_SYNC) \
-		build-gateway build-orchestrator build-sandbox build-litellm
+		build-gateway build-orchestrator build-sandbox build-litellm build-redis
 
 # Per-image sub-targets so `build` can run them under -j. They assume the
 # repo-deps/ build context has been prepared (the `build` target does that
@@ -543,6 +543,9 @@ build-sandbox:
 
 build-litellm:
 	DOCKER_BUILDKIT=1 docker build $(call image_tags,egg-litellm) -f config/litellm/Dockerfile config/litellm
+
+build-redis:
+	DOCKER_BUILDKIT=1 docker build $(call image_tags,egg-redis) -f config/redis/Dockerfile config/redis
 
 # ============================================================================
 # Kubernetes (k3s) targets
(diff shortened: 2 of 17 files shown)


jwbron:
**Live-canary finding, fixed in 2b2dc9f40.** Deployed this branch to the local cluster and submitted a throwaway pipeline as the first real traffic through the Redis backend. Writes and non-blocking reads worked immediately (stream keys created, heartbeats flowing), but every agent long-poll errored at exactly 5s:

```
ERROR orchestrator.redis_message_store: Failed to read from Redis Stream pipeline_id=pipeline-e73d8284 error="Timeout reading from socket"
```

Root cause: the connection pool sets `socket_timeout=5`, and redis-py enforces that on the blocked read itself — so a single `XREAD BLOCK 25000` (the standard agent wait) dies at the 5s mark and the route 500s. Agents still made progress because the wait CLI retries, but event-driven waiting was degraded to error-spam polling. This is precisely the class of bug the issue predicted (`fakeredis` has no sockets, so unit tier structurally cannot catch it) and the reason for deploying real Redis in CI.

Fix: blocking reads are sliced into `XREAD BLOCK` chunks of at most `_MAX_BLOCK_MS` (4s, under the socket timeout) looped against the caller's wait deadline — data still returns immediately, idle slices cost one extra round-trip each. A `TimeoutError` that still fires on a blocking slice degrades to an idle slice; non-blocking reads keep raising. Five new unit tests pin the cap (including its `< socket_timeout` relationship), both blocking branches, and the degradation scoping.

Re-running the canary (including a mid-phase orchestrator restart to exercise the #3076 durability story) after redeploying with this fix.

jwbron:
**Round-2 canary + restart drill: PASS** (cluster redeployed with 2b2dc9f40 + 887b3ef25).

- **Long-poll fix verified**: 12+ minutes of wrapper `wait-loop` long-polls (`WAITING_FOR_EVENT` heartbeat cycles), **0** `Failed to read from Redis Stream` errors. Round 1 produced 8 errors in the first 2 minutes at the same stage.
- **#3076 durability drill**: restarted the orchestrator mid-refine with 20 messages of live transcript. The stream not only survived but grew to 30 during the orchestrator's downtime — agents kept writing to Redis while the coordinator was gone. Startup reconciliation recognized the pipeline (`pipeline has live pods, leaving RUNNING`), and agent wrappers resumed against the new pod within ~9s of boot (`consensus/next-action` and heartbeats all 200). No fallback marker, no read errors on the new pod.
- Canary cancelled with cleanup after the drill; no artifacts left behind.

On the old in-memory backend this restart would have dropped the entire transcript and de-synced the consensus tracker. End-to-end validation of both halves of the issue (live Redis coverage + restart durability) is done.

jwbron:
**Full live-validation record for this branch** (consolidating the thread):

1. **Deploy verification** — `make redeploy` builds/publishes `egg-redis` through the loopback registry alongside the other images; redis pod Running with `redis-data` PVC Bound (1Gi local-path); `await-egg-deploy.sh` gates on the redis Deployment; `/api/v1/health` reports `message_store: {status: ok}`. All 3 new integration tests pass against the live cluster, including the in-pod XADD/XRANGE/DEL round-trip through the production selection path (`BACKEND RedisMessageStore`).
2. **Round-1 canary** — caught the `XREAD BLOCK` vs `socket_timeout=5` bug within 2 minutes of real traffic (see above); fixed in `2b2dc9f40` with 5 unit tests pinning the slice-cap contract.
3. **Round-2 canary + restart drill** — 0 read errors over 12+ min of wrapper long-polls; mid-phase orchestrator restart survived with the transcript growing 20→30 in Redis *during* the downtime; wrappers resumed in ~9s (see above).

Follow-ups spawned by this validation, both out of scope here:
- #3154 — image-sourcing policy (local wrapper vs digest-pinned public references), revisit on the second third-party image or GKE (#3002).
- #3157 — codify the event-pump waiting contract (agents never block on the bus; the wrapper owns the wait) and retire the stale STAY ALIVE canon — the docs confusion surfaced while attributing the round-1 long-poll errors.

Issue #3288, opened 2026-06-26
Documenter agent + docs: snapshot of current state, not a ledger of changes


jwbron:
## Problem

Documentation in egg has drifted into acting as a **ledger of changes** rather than a **snapshot of current state**. This is most pronounced in artifacts produced during the SDLC pipeline, where the documenter agent (and, by accretion, the doc corpus) records *process* context — which slice landed a change, which TASK-N implemented it, what the code "used to" do, what was removed in issue #NNNN — interleaved with the description of how the system actually works today.

The pipeline-process framing (slices, tasks, HITL iteration numbers, per-phase rollout order) is meaningful *while a pipeline runs*. Once the pipeline completes and the change is on `main`, that ledger is noise: it does not describe the current codebase, it costs the reader implementation-roadmap knowledge to parse, and it steadily rots as later changes invalidate the narrative.

We want documentation to be a **snapshot of the current state of the codebase**. Historical context should appear only where it is *tangibly valuable* to a reader of the current system (e.g. "this is intentionally email-allowlist-based because impersonation is the threat model"), never as a default record of how the work was sequenced. In particular, **SDLC "slice" context must never appear in inline docs, docstrings, or reference documentation.**

## Evidence / scope

This is widespread. A survey found slice/TASK/issue-as-history references in roughly 8–12% of sentences across core architecture pages, plus heavy inline-comment and docstring tagging across `orchestrator/`, `gateway/`, and `shared/`. Representative examples:

**Architecture docs reading as change-logs:**
- `docs/architecture/brc-memory.md` — "Introduced by #2908, slice-1.", "Slice-1 lands the writer… lands in slice-3. Slice-4 flipped the default…", "Added in #3200 slice-5."
- `docs/architecture/orchestrator.md` — behavioral sections keyed on "slice-4 TASK-4-5", "TASK-4-1 … used to clear only…"; a whole "slice lifecycle" section narrating slice-1→slice-4 rollout as a change log.
- `docs/architecture/slice-dag.md` — "The implement phase used to run as a single monolithic agent team… The slice-DAG model replaces that…"
- `docs/architecture/gateway-auto-filter.md` — sections explicitly "preserved for context" describing a "historical #1882 design" whose code has been removed; a "## What was removed" section.
- `docs/architecture/coordination-state.md` — "This page describes the final shape of the #3077 invariant. The epic was landed in six slices, all of which have now shipped…"; Threat/Mitigation/Status tables full of "Retired surface", "now historical".

[...]

Issue #2548, opened 2026-05-07
Slice PRs are missing analysis/plan docs and all BRC history; need a 'context' PR for refine+plan phases plus per-slice BRC in each slice PR


jwbron:
## Observation

The slice-1 PR for pipeline \`issue-2474-v2\` is [#2533](https://github.com/jwbron/egg/pull/2533). Looking at the file list and the body, the PR is missing:

1. **The refine-phase analysis document** (\`.egg-state/drafts/2474-analysis.md\`) — committed on \`egg/issue-2474-v2/work\` at commit \`0c92bab7d\` during refine. Reviewers approaching the slice-1 PR have no link to the problem statement or recommended approach that motivated the change.
2. **The plan-phase plan document** (\`.egg-state/drafts/2474-plan.md\`) — committed on \`egg/issue-2474-v2/work\` at commit \`6d325e9f4\` during plan. Reviewers can't see how the work was decomposed into slices, what defaults the planner chose for HITL questions, what risks were flagged, etc.
3. **The refine-phase BRC history** (\`.egg-state/brc-history/2474-refine.json\` + \`.md\`) — the proposal/ACK/CONFIRM record from the refine BRC round. Establishes that reviewer_refine and reviewer_agent_design ACKed the analysis.
4. **The plan-phase BRC history** (\`.egg-state/brc-history/2474-plan.json\` + \`.md\`) — same for the plan BRC round.
5. **The slice-1 implement-phase BRC history** — the proposal/NACK/ACK/CONFIRM record specific to slice-1. The reviewer_code NACK on coder v1 (which caught the \`.github-staging/\` retirement-stub mistake), the reasoning behind the v1→v2→v3 progression, and the consensus close. None of this is visible to a PR reviewer.

The slice-1 branch (\`egg/issue-2474-v2/slice-1\`) is based on \`egg/issue-2474-v2/work\` (which carries the analysis and plan files), so technically those files exist somewhere upstream. But \`origin/main\` is the merge target — the analysis/plan files are NOT in slice-1 PR's diff against \`main\` because they only exist on \`egg/issue-2474-v2/work\` and that branch's commits aren't part of slice-1 PR's review surface. They'll never be reviewed, never merge.

## Proposed architecture

Restructure the PR train so reviewers see context up-front:

### \"Context\" / \"preface\" PR

A new PR (could be \"slice-0\" or named explicitly) that ships:
- \`.egg-state/drafts/<id>-analysis.md\` (refine output)
[...]

PR #3333, opened 2026-06-26
Advisory ledger-reference ratchet lint (#3328)

### LINKED ISSUE #3328: Advisory lint/CI ratchet against new slice-N/TASK-N ledger refs in docs (follow-up to #3288, cq-2)
## Follow-up to #3288 (durability guardrail, per HITL cq-2)

#3288 reframed the documenter to stop emitting SDLC ledger refs and cleaned the existing corpus, but — by explicit operator decision (cq-2 = "no — out of scope, note as follow-up") — did **not** add an automated guard to keep `slice-N` / `TASK-N` / `cq-N` ledger references from creeping back into docs/docstrings later. This issue is that deferred guard.

### Proposal
Add a lightweight **advisory lint/CI ratchet** that flags when a committed doc/docstring/comment introduces a **new** ledger reference (`slice-N`, `TASK-N`, `cq-N`, "what was removed in #NNNN" style change-log narration) — ratchet semantics (don't fail on the pre-existing deferred tail, only on net-new), so it doesn't block on the long tail tracked separately.

### Must avoid false positives on live-machinery vocabulary
The hard part (why it was deferred — it has its own tuning cost) is distinguishing **ledger** refs from **live runtime** uses that legitimately contain the same tokens:
- `docs/templates/plan.md` `TASK-N` = the live plan format.
- The slice-DAG domain vocabulary (`slice_id`, `contract.slices`, `EGG_*_SLICES`, per-slice Job names).
- Contract-task-id fixtures/handling in `.py` (`task-N-N` as the live id format).

Scope it as advisory first (warn, don't block) and tune the allowlist before considering a hard gate.

jwbron:
Closes #3328. Follow-up to #3288 (durability guardrail deferred by HITL cq-2).

## What

#3288 de-ledgered the docs corpus and reframed the documenter to emit snapshots, but — by explicit operator decision (cq-2 = "out of scope, note as follow-up") — did **not** add an automated guard to keep `slice-N` / `TASK-N` / `cq-N` ledger references from creeping back in. This adds that deferred guard as an **advisory** (warn, never block) CI ratchet.

## How

New custom check `scripts/check-ledger-references.py`, auto-discovered by `make lint-custom` (any `scripts/check-*.py`):

- **Detects** the three unambiguous SDLC ledger token classes — `slice-N`, `TASK-N`, `cq-N` — across repo markdown plus non-test Python source (docstrings/comments).
- **Ratchet semantics** via `scripts/ledger-references-baseline.yaml` (per-file counts): flags only **net-new** tokens, so the long pre-existing tail (tracked separately, decaying as docs are touched) does not fire on every run. Lowering a count never warns. Re-snapshot with `--update-baseline`.
- **Advisory by default**: prints warnings and exits `0` so it never blocks a PR. `--strict` makes net-new a hard failure (intentionally *not* wired into CI yet — per the issue's "advisory first, tune the allowlist before a hard gate").

### False-positive controls (the reason the guard was deferred)

The issue calls out live-runtime vocabulary that shares these tokens. Handled by:
- **Hyphenated patterns** skip live identifiers: `slice_id`, `contract.slices`, `EGG_*_SLICES`, "the slice DAG" all use underscores / dotted access / bare words and don't match.
- **Structural exclusions**: `docs/templates/` (`plan.md`'s `TASK-N` is the live plan format), `.egg-state/` (pipeline state + BRC transcripts are a ledger by design), and test files (fixtures use `slice-N` / `task-N-N` as live data).
- **Per-line escape hatch**: `ledger-ok` in a comment on the line excludes it (e.g. a legitimate new slice-DAG error message).

### Deliberately out of scope for v1

Change-log *prose* ("what was removed", "used to … now …") — it can't be regex-matched without high false positives against the issue links that legitimately justify *why* the current system is shaped the way it is (which #3288 explicitly keeps). Noted in the script docstring as a tuning follow-up.

## Tests

`tests/scripts/test_check_ledger_references.py` (30 tests): pattern match/non-match, occurrence counting, suppress marker, exclusions, ratchet evaluate (over/at/under baseline, new file), baseline round-trip, and a guard that the committed baseline is clean against the live corpus (so a fresh checkout produces no spurious advisory noise).

- `make lint-custom` → `ledger-references` runs clean.
- `ruff check` / `ruff format --check` / `mypy` clean on new files.


jwbron:
Pushed `e5ceb22c8` to re-scope this PR and unblock CI.

**What changed**
- **Reverted the bundled #2548 doc edits** (`concurrent-execution.md`, `orchestrator-cli.md`). Those two doc-test failures (`test_cross_references_issue_2548`, `test_slice_1_paragraph_ties_to_context_branch`) are red on `main` independently of this PR and are owned by the separate, green **#3325**. Keeping the edits here duplicated #3325 and would conflict on whichever lands second.
- **Refreshed the baseline** so the ratchet's own self-test (`test_no_net_new_against_committed_baseline`) stays green in *both* merge orderings: `concurrent-execution.md` goes 15→16, matching #3325's single added `slice-1` token. With the baseline at the post-#3325 count and this PR's tree doc-free, net-new is empty whether or not #3325 has landed.

This PR is now scoped to its three ledger files (`scripts/check-ledger-references.py`, `scripts/ledger-references-baseline.yaml`, `tests/scripts/test_check_ledger_references.py`).

**Merge order:** land **#3325 first** (it clears the #2548 doc tests on `main`), then this PR's merge tree goes fully green.

**Caveat (design):** `test_no_net_new_against_committed_baseline` checks the committed baseline against the *live merge corpus*, so it will re-flag whenever a later PR brings new `slice-N`/`TASK-N`/`cq-N` tokens from `main` — a maintenance treadmill inside the *blocking* unit suite, somewhat at odds with the issue's "advisory, never block" intent. Worth considering moving that self-test out of the blocking tier (keep it as a `--update-baseline` reminder in the advisory script).

PR #3457, opened 2026-07-03
Fix Squid localnet ACL for Kubernetes pod CIDRs

jwbron:
## Summary

`gateway/squid.conf` defines `acl localnet src 172.32.0.0/24` — the Docker Compose egg-isolated subnet. In the Kubernetes deployment, clients reach Squid with their pod IPs (Cilium cluster-pool / k3s ranges under `10.0.0.0/8`), so **no client has ever matched the allow rules**: every proxied CONNECT was accepted for SNI peek, denied post-peek by `http_access deny all`, and bumped with Squid's self-signed certificate instead of tunneled.

This has been latent since the k8s migration because all first-class traffic bypasses the proxy — Anthropic API via the gateway's credential-injecting endpoint, git/gh via the REST wrappers, Jira/Confluence via `/api/v1/*`. The egress-allowlist path (#3455) is the first consumer that actually needs Squid to tunnel, which is how this surfaced.

## Diagnosis evidence

- `curl -x http://gateway:3129 https://registry.npmjs.org/...` from an in-cluster pod fails with `SSL certificate problem: self-signed certificate in certificate chain` — the bumped-denial signature, not a tunnel.
- Squid access log shows `CONNECT` entries with `status: 200, bytes: 0` (the "Connection established" reply that precedes the post-peek denial).
- Client IPs in the log are `10.0.0.x`; nothing in `172.32.0.0/24` exists in the cluster.

## Change

One line: `acl localnet src 172.32.0.0/24 10.0.0.0/8` (+ explanatory comment). Port 3129 reachability is already gated by NetworkPolicy (agent pods → gateway only), so the ACL is defense-in-depth; the domain allowlist continues to constrain destinations.

## Deployment note

Baked into the gateway image — takes effect on the next gateway rebuild + rollout.

### DIFF
--- a/gateway/squid-allow-all.conf
+++ b/gateway/squid-allow-all.conf
@@ -36,8 +36,15 @@ http_port 3129 ssl-bump \
 # Access Control Lists
 # ==============================================================================
 
-# Define local network (egg-isolated subnet)
-acl localnet src 172.32.0.0/24
+# Define local network: the Docker Compose egg-isolated subnet plus the
+# Kubernetes pod range. In k8s, agent pods reach Squid with their pod IP
+# (Cilium cluster-pool / k3s CIDRs live under 10.0.0.0/8); without that
+# range here, every proxied CONNECT fails http_access and gets bumped
+# with Squid's self-signed certificate instead of tunneled. The full /8
+# (rather than a tighter k3s/Cilium subrange) is deliberate: it covers
+# both the k3s pod/service CIDRs (10.42/10.43) and the Cilium cluster-pool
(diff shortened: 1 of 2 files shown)


jwbron:
Added a second commit: `dns_nameservers 127.0.0.11` hardcodes Docker's embedded DNS, which doesn't exist in a k8s pod — even with the localnet fix, Squid couldn't resolve upstream hosts to build the tunnel. Dropped the directive so Squid uses the container's `/etc/resolv.conf` (same embedded resolver under Compose, cluster DNS under k8s).

Issue #3398, opened 2026-06-30
Pipeline ships red slice PRs: lint drift (pre-commit bypassed), slicing-order test failures, dangling gaps, add/add contract conflicts


jwbron:
## Summary

Every PR produced by the `issue-2270-overhaul` pipeline opened **red** — the 9 slice PRs (#3282–#3297) and the context PR #3280. The failures were not deep logic bugs; they were process/wiring gaps that should never have reached an opened PR. **Newly-opened slice PRs should be green on open** (or at minimum each independently lint-clean and CI-passing). This issue tracks figuring out which of the observed failure classes are *structural* (slicing/planning/reconciliation) vs *misconfigured local toolchain* and fixing them so the pipeline stops shipping red PRs.

Concrete instance: stack rooted at #3280 (`egg/issue-2270-overhaul/work`→main), slices #3282(s1) … #3297(s9).

### 1. ruff-format drift on every slice — **looks like a toolchain/config gap (most actionable)**
Every slice failed the `Python` lint check on `ruff format --check` (CI uses `ruff==0.15.12`, pinned `ruff>=0.15.12,<0.16` in `pyproject.toml`). The agents' committed code was wrapped differently than 0.15.12 formats it.

Root cause found: the agent commit path **bypasses pre-commit**. `gateway/git_client.py:322` and `:689` commit with `--no-verify`; `gateway/commit_observer.py:12` documents "decision-1(d): agents can suppress hooks with `--no-verify`". `.pre-commit-config.yaml` *does* pin `ruff` + `ruff-format` at `v0.15.12` — but the hook never runs on agent commits, and the sandbox image doesn't appear to install pre-commit at all. So nothing runs `ruff format` before the slice PR is opened, and any formatter drift (or use of a different ruff in-sandbox) sails straight to CI.

→ Decide the intended contract: should the producer run `make lint-fix`/`ruff format` (pinned version) as a gate before opening a PR, or should the commit path stop using `--no-verify` for the format hook, or should a server-side autofix run? Whatever it is, the slice PR should not open with format drift.

### 2. Latent mypy failure masked by lint ordering — **structural (new test root added without full wiring)**
[...]

PR #3517, opened 2026-07-06
Green gate Stage A: config-driven format autofix before slice PR open

### LINKED ISSUE #3409: Green gate Stage A: server-side format autofix before slice PR open
Follow-up to #3398 (per-slice green gate). Stage B (run configured checks at the slice tip, block red PRs) lands there; this issue carries **Stage A: config-driven format autofix** at the gate, deferred because #3299 (sandbox ruff pinned to the CI constraint) already neutralized the formatter-drift half of failure class 1.

## Scope

When the green gate's `lint` check fails at the slice tip:

1. Add an optional per-repo `fix` command to the `checks` schema in `repositories.yaml` (`config/repo_config.py::validate_checks` currently accepts exactly `{name, command}`) — e.g. egg's `lint` check gets `fix: make lint-fix`.
2. The check-runner executes the configured `fix` command and returns the resulting patch in its verdict.
3. If the patch is non-empty, the **orchestrator** applies and commits+pushes it to the slice integration branch **via the gateway** (the sanctioned writer — the runner never pushes), attributed to a system/orchestrator identity (precedent: on-merge-conflict CI bot #3255, git-route attribution #2919), then re-runs the check.
4. No `fix` configured ⇒ lint failures route back to the slice team like any other red (Stage B behavior).

## Notes

- The gateway push authorization likely already admits this: `_SLICE_INTEGRATION_BRANCH_RE` + synthetic-session exemption (`gateway/gateway.py`) covers orchestrator pushes to `egg/<id>/slice-N`. Verify the role+phase path end-to-end before relying on it.
- Must land such that dependent slices fork *after* the format commit, or they inherit unformatted code and re-trip the gate.
- Residual class-1 risk this covers: "tester didn't run/finish the format step" (toolchain drift itself is dead via #3299).

jwbron:
Closes #3409. Follow-up to #3398 (Stage B landed in `orchestrator/slice_green_gate.py`).

## What this does

When the green gate finds a configured check red at the slice tip and that check carries an optional `fix` command in `repositories.yaml` (e.g. a `lint` check with `fix: make lint-fix`):

1. The check-runner executes the fix inside its worktree, re-runs the check, and reports a `fix` sub-object in the verdict (fix exit code, re-run result, changed tracked files). The check's `ok` stays false: the tip as pushed is still red.
2. If every failed check re-ran green, the orchestrator stages the tracked modifications (`git add -u`, so untracked check droppings like caches never enter the commit), commits as `egg-green-gate` (`--no-verify`, mirroring the `egg-salvage` system-identity precedent and #2919 orchestrator attribution), and pushes to the slice integration branch via the launcher-authed gateway push route (`push_worktree_branch`). The runner never pushes.
3. On push success the gate passes and the slice closes. Any autofix failure (nothing to commit, commit error, push rejection) blocks the slice exactly like an unfixed red, with the error appended to the failure message.
4. Checks without `fix`, or whose re-run stays red, or a verdict where only some failed checks are fixable, route to the slice team unchanged (Stage B behavior).

## Design notes (deviations from the issue sketch, both simplifications)

- **No patch transport.** The issue sketch had the runner return a patch that the orchestrator applies. The runner pod mounts the gateway worktree via hostPath, so the fix's modifications are already on disk when the pod exits; the orchestrator stages and commits from that same worktree. The verdict carries the changed-file list (capped at 100 entries) for observability only.
- **No second runner pass.** The re-run happens inside the runner against the exact tree that gets committed, with the repo's pinned toolchain, so the committed tip is already validated green.
- **Gateway authorization**: the orchestrator push uses the launcher-auth path (`/api/v1/git/push` with `use_launcher_auth=True`), which the gateway treats as orchestrator-trusted; the `_SLICE_INTEGRATION_BRANCH_RE` + synthetic-session exemption the issue mentions is the session-token path and is not needed here.
- **Fork ordering**: the push lands before any close side effect, and dependent slices resolve their parent tip via `git ls-remote` on origin when they start (`create_slice_integration_branch`), so they fork after the format commit and cannot inherit unformatted code.
- **Rollout**: the autofix commits/pushes only in `on` mode. In `log` mode the runner still executes the fix (soak signal: the log line carries `autofix_ready`), but nothing is committed or pushed.

## Changes

- `shared/egg_config/validators.py::validate_checks` (+ the two import-fallback copies in `config/repo_config.py` and `orchestrator/routes/pipelines/__init__.py`): accept and preserve an optional `fix` key; empty/None `fix` is dropped; unknown keys still filtered.
- `orchestrator/slice_green_gate.py`: runner fix execution + verdict extension; `_commit_and_push_autofix`; gate wiring + docstrings.
- `config/repositories.yaml.example`: document the `fix` knob. The live `repositories.yaml` is deployment-side; egg's `lint` check needs `fix: make lint-fix` added there for this to take effect.
- Tests: `orchestrator/tests/test_slice_green_gate.py` (runner fix flow executed for real in a subprocess with a real git repo; `_commit_and_push_autofix` against a real repo with a mocked gateway; gate wiring for fixed/unfixed/partial/log-mode paths), `tests/egg_config/test_validators.py` (new `TestValidateChecks`).

## Testing

- `orchestrator/tests/test_slice_green_gate.py` + `tests/egg_config/test_validators.py`: 134 passed.
- Pre-commit (ruff, ruff format, whitespace hooks) clean.


jwbron:
Refreshed against `main` and folded in the stranded review-fix commit.

- Cherry-picked `980d1fb15` ("Green gate autofix: disable hooks + gate on final full re-run"), which was sitting on `egg/issue-3409-green-gate-autofix-review-fixes-v2` and had never been pointed at this PR. It disables git hooks on the orchestrator's autofix commit (`-c core.hooksPath=/dev/null`; `--no-verify` alone left post-commit and `git add` hooks free to run agent-controlled worktree hooks on the orchestrator host) and makes "the committed tip is validated green" true in general by adding a final full re-run plus an untracked-files check.
- Merged `origin/main` (the branch was 419 commits behind; no conflicts).
- `orchestrator/tests/test_slice_green_gate.py` + `tests/egg_config/test_validators.py`: 147 passed. Ruff clean.

Sequencing note: this is throughput/toil reduction, not a correctness prerequisite for the gate. It also widens the trust surface (the orchestrator starts committing and pushing to the slice integration branch as `egg-green-gate`), and the autofix only commits in `on` mode. Suggest landing it after the gate has actually soaked in `log` mode.

Issue #3548, opened 2026-07-07
BRC ack accepted but not visible in slice tracker; no-op-parked rounds stall silently; restart_agent delegated respawn never fires


jwbron:
## Summary

During slice-8 of pipeline issue-3523 (2026-07-07, after the #3544 deploy), the BRC round stalled for ~30 minutes in a shape where reviewer verdicts were accepted by the ack API but never became visible in the slice consensus tracker, the event loop then absorbed the whole round into no-op parks with zero operator signal, and `restart_agent`'s delegated respawn never fired. The round only converged after a full `restart_phase` re-round.

## Timeline (UTC)

- 21:07:08 documenter proposes (SHA `d4df6c734`).
- 21:07:15 reviewer_code and reviewer_contract spawn (action=ack), run ~2 min each, exit 0. Both agents' transcripts state they posted ACKs and received `ok: true, fully_acked: true`.
- 21:09 onward: `get_consensus_status` shows both reviewers stuck in `reviewer_phase: REVIEWING`, 0 confirmed, no NACKs. No `CONSENSUS_ACK` ever appears on the message bus. The other three reviewers are never spawned.
- 21:09-21:36: every spawn arm is no-op-parked; the event loop goes fully silent (no spawns, no alerts, `pending_decisions` empty). The overseer pod is reaped at 21:26 without escalating.
- ~21:36 `restart_agent(reviewer_code, slice-8)` returns "restarted successfully" with `container_id: ""` and "respawn delegated to event loop"; no container is ever spawned for it (reproduces the same behavior seen at 18:57 with the documenter).
- 21:41 `restart_phase(implement)` re-rounds. Documenter re-proposes; reviewer_contract re-asserts its ACK at 21:43:09 ("My prior ACK returned fully_acked: true, but the matrix still shows reviewer_contract as WORKING and the event re-presents this review as pending, so the edge may not have registered. Re-asserting the ACK idempotently."). At 21:44 the tracker STILL showed REVIEWING for it.
- 21:48:34 the round nevertheless converges (all 8 CONFIRMED) and the slice completes.

## Defects to extract

[...]

Issue #3520, opened 2026-07-06
agent-invocation-noop-streak fires high-priority for a role that is WAITING_ON_ROLE on a live producer


jwbron:
## Observed

Pipeline `issue-3364` (2026-07-06), twice in ~30 minutes:

- refine, 18:18:45Z: `agent-invocation-noop-streak: event-loop [high]` for role=simplifier, streak=3 (then re-alerted at streak=4 at ~18:21Z), dedupe key `2f156e27...`
- plan, 18:44Z: same anomaly for role=simplifier, streak=3, dedupe key `19102154...`

In both cases the alert was a false positive in the operator-facing sense:

- The simplifier's own HEARTBEATs during the streak window carried `metadata.state = WAITING_ON_ROLE` with `metadata.waiting_on = refiner` (refine) / the plan producers (plan), plus an explicit body: "blocked on refiner CONSENSUS_PROPOSE ... Will propose once refiner proposes", including its own escalation threshold.
- The waited-on producer was live and heartbeating the whole time.
- The park mechanism itself worked exactly as designed: the arm un-parked as soon as the refiner's CONSENSUS_PROPOSE landed (BRC movement) and the simplifier proposed and confirmed in the same cycle.
- The refine-phase alert took the empty-fingerprint branch ("No unresolved contract decision was visible at park time"), i.e. the code itself knew its primary hypothesis (gating cq-N) did not hold.

## Why it matters

`_emit_noop_alert` (orchestrator/event_loop/_supervisor.py:578, added in #3425) was designed for the wedged-on-unresolved-cq-N case and is priority=high unconditionally. A dependent role no-oping while it waits for its upstream producer's first proposal is normal BRC choreography in every phase that has the simplifier (or any consumer role), so this alert fires as a matter of routine early in refine and plan. That trains operators to skim past high-priority alerts; #3364's whole visibility thread is about making alerts trustworthy, and item 9 of the #3364 addendum (alert feed with severity and dedup) is directly undermined by a routine false positive at [high].

Parking the arm is fine here (it saves pod spawns); only the alert severity/emission is wrong.

## Proposed fix

In the supervisor, before emitting the noop-streak alert with priority=high, consult the parked role's most recent HEARTBEAT metadata on the bus (message_store already carries `state`/`waiting_on`; see WAITING_ON_ROLE usage in orchestrator/message_store.py):

[...]

Issue #3478, opened 2026-07-03
Event loop spawns a producer whose in-slice dependencies are unmet, burning the no-op streak and parking the arm


jwbron:
Observed on image 098bd6acd, single-node k3s, slice-DAG implement phase.

## Symptom

A slice has 3 coder tasks and 1 tester task, where the tester task (a test suite for the coder's modules) cannot start until the coder artifacts exist. The event loop derived action=propose for the tester and spawned it 3 times while the coder was still producing. Each invocation exited cleanly as a no-op (the agent itself correctly reported it was blocked), which tripped the JobSupervisor no-op streak and parked the tester arm:

```
19:33:04 orchestrator.progress: Progress event emitted agent_role=tester step="task-9-3 smoke.test.ts (CI gate): blocked on coder artifacts" state=blocked
19:33:29 orchestrator.event_loop: event-loop spawn dispatched slice_id=slice-9 role=tester action=propose dedupe_key=c6b3c2d9...
19:34:37 orchestrator.event_loop: event-loop spawn dispatched slice_id=slice-9 role=tester action=propose dedupe_key=c6b3c2d9...
19:35:20 orchestrator.event_loop: JobSupervisor: 3 consecutive successful no-op invocations for key=c6b3c2d9... (action=propose, role=tester): parking the arm; spawning at an unchanged BRC state cannot resolve an operator-bound wedge
```

The agent even emitted a progress event saying it is blocked on the other producer's artifacts, but the derivation kept re-deriving propose for it at an unchanged BRC state.

## Impact

Not a wedge (the #3467 un-park on BRC movement plus the 1800s probe should recover it once the coder proposes), but it wastes 3 pod spawns per slice with this shape and can add up to 30 minutes of latency if the un-park signal is missed. Every slice whose tester depends on same-slice coder output hits this.

## Suggested direction

When deriving next-action for a producer, check whether its pending task rows depend on artifacts owned by another producer in the same slice whose rows are still pending; derive wait (not propose) until those rows complete or the depended-on producer proposes. Alternatively: treat an agent-reported blocked progress event as a park-without-streak signal so the arm parks immediately (0 wasted spawns) and un-parks on the dependency's propose.

jwbron:
[...]

Issue #3425, opened 2026-07-02
[orchestrator/event-loop] Unbounded agent re-spawns against a slice wedged on unresolved HITL (cq-N): successful no-op invocations never park (~50 spawns observed)


jwbron:
## Summary

When a slice's next BRC action is blocked on an **unresolved operator HITL decision** (contract `cq-N`), the event loop keeps re-deriving the same action and re-spawning one-shot agent pods indefinitely — each pod starts, discovers the block, and exits as a successful no-op, so nothing changes and the loop goes again. Observed: **~50 consecutive coder re-spawns with zero BRC progress** on one slice over ~50 minutes.

The #3138 fast-fail escalation does not catch this: its streak-park keys on consecutive agent-invocation **failures**, and these invocations **succeed** (the agent runs and exits cleanly without proposing). So there is no park, no backoff — just an unbounded spawn-burn until the human answers.

## Concrete instance

Pipeline `issue-3393`, slice-4, 2026-07-02 (~06:20–07:15Z):

1. reviewer_contract registered HITL `cq-3` (a plan-scope gap) and withheld its ACK pending the operator's answer; all other reviewers CONFIRMED. Slice-4 is legitimately wedged on the operator.
2. The event loop derived `action=propose` for `role=coder`, spawned a one-shot pod; the coder had nothing actionable (blocked on cq-3), exited without a bus message.
3. No BRC state change → same derived action → new spawn. Repeat.
4. Overseer alerts fired correctly (`pending 1818s without BRC-bus progress`, then an escalation noting "~50 consecutive coder one-shot re-spawns... ZERO progress"), so *detection* worked — but detection doesn't stop the spawn loop; only the HITL resolution does.

## Why it matters

- Each cycle costs a pod spawn + a full agent bootstrap + at least one model invocation — pure waste while a human is (reasonably) away for an hour. Overnight, this is hundreds of pointless spawns per wedged slice.
- The overseer alert stream during the burn is noise layered on a known, already-alerted condition.

## Expected behavior (suggestion, pending root-cause review)

When the event loop derives an action for a slice that has an **unresolved contract HITL decision gating it** (or more generally: when N consecutive *successful-no-op* invocations produce zero bus progress for the same dedupe key), it should park that arm exactly like the #3138 failure-streak park — sticky, alert once, resume automatically when the BRC state changes (e.g. `decision.resolved` for the gating `cq-N`). The wedge is operator-bound; spawning agents at it cannot resolve it.

## Related

[...]

PR #3607, opened 2026-07-25
Gate the simplifier's first propose on its upstream producer (#3603)

### LINKED ISSUE #3603: noop-streak park fires on correctly-blocked agents and can starve the only path to progress
## Summary

`agent-invocation-noop-streak` parks an agent arm after 3 clean exits with no BRC progress. It cannot distinguish "stuck" from "correctly blocked waiting on an upstream producer", which is the normal opening state of every BRC phase.

## Evidence

Reproduced 4/4 across the assessment series, prompt-independent AND phase-independent: at 5, ~13 and ~14 minutes into three different refine phases, and ~3 minutes into a plan phase.

In the plan-phase instance the upstream producer was demonstrably healthy (diverse tool use, `mcp__brc__propose` already emitted). The parked role reported `WAITING_ON_ROLE` with the blocking role named in `waiting_on` metadata on every heartbeat, so the information needed to distinguish the two cases was already present and ignored.

Separately, in `issue-3596-v2` implement slice-1 the park became load-bearing: it parked the coder, which was the sole remaining producer, and the event loop reported it "cannot advance: every derivable spawn arm (coder/propose) is blocked on a no-op-parked dedupe key and no one-shot Job is in flight".

## Why it matters

Normally this is noise (wasted spawns, misleading [high] alerts pointing at the correctly-behaving agent). But when the parked arm is the only remaining producer it converts a heuristic into a stop, and the 1800s probe backstop cannot converge if the underlying cause is unchanged.

## Suggested fix

1. Do not park a role whose heartbeat names an unsatisfied upstream producer in `waiting_on`.
2. Never park an arm that is the sole remaining producer for the slice; or force an immediate probe plus escalation when the event loop detects every arm blocked, rather than waiting out the 1800s backstop.

## Related

#3595 root cause 2, #3601

jwbron:
## What

Makes the #3478 first-propose gate honour `wake_only` edges, so the de-roled `simplifier` derives `wait` until its upstream producer proposes instead of polling into the #3425 no-op park.

Fixes the reproduced half of #3603 (= #3595 root cause 2). See "What this does not do" for the rest.

## Why

The park's defect, stated precisely: **repeated clean no-op exits are the simplifier's specified wake mechanism, and the park counts exactly those exits.** Two subsystems contradict each other and the park wins.

- `ReviewEdge.wake_only` (`review_graph.py`) documented the wake as "a WORKING producer re-derives `propose` on every event-loop poll, and a clean orient-and-exit ... frees the spawn-dedupe key, so the orchestrator re-spawns it until it proposes."
- `JobSupervisor.record_success` (#3425) parks an arm after 3 clean completions with no BRC progress and fires `agent-invocation-noop-streak [high]`.
- The simplifier's own prompt (`_prompt_phase.py`) tells it: *"you begin writing only once that producer issues `CONSENSUS_PROPOSE` ... Do not race ahead before the draft exists."*

So the agent obeys its prompt, the derivation spawns it anyway, and the supervisor parks it for obeying. #3595 reproduced this **4/4, prompt-independent and phase-independent** (5, ~13, ~14 min into three refine phases; ~3 min into a plan phase). The wedge precondition is satisfied by correct behaviour on the common path — the opening state of every BRC phase.

#3478 already built the right mechanism for this exact shape (`tester` gated on `coder`, `risk_analyst` gated on `architect` + `task_planner`) and its docstring names the failure it prevents: "three such no-ops burn the #3425 streak and park the arm". It exempted wake-only edges, so the one role whose entire job is "wait for the upstream draft" was the one role left polling.

## The change

`wake_only` means **"casts no verdict", not "no dependency"**. The edge exists precisely because the companion is written against the upstream producer's draft. `_pre_propose_upstream_producers` now includes wake-only edges in the upstream set:

| Phase | Role | Gated on |
|---|---|---|
| refine | `simplifier` | `refiner` |
| plan | `simplifier` | `task_planner` |
| plan | `risk_analyst` | `architect` + `task_planner` (unchanged) |
| implement | `tester` | `coder` (unchanged) |

The wake becomes level-triggered on the derivation: the first poll that sees the upstream's proposal derives `propose` and spawns the arm. No spawn before that → no no-op → no streak → no park → no mis-targeted alert.

Everything else about `wake_only` is untouched — still excluded from `_has_pending_peer_proposals` and the reviewer confirm guards, so the simplifier is never assigned a spawn-able `ack` it cannot satisfy (the #3381 regression) and still confirms without ever voting. Pinned by `test_next_action_wake_only_edge_still_casts_no_verdict`.

**Acyclicity guard fixed alongside.** Terminality is now judged over the same edge set the upstream computation uses. Judging it over review-bearing edges only would call each half of a mutual wake-only pair terminal, gate both, and wedge the phase with no producer able to move — a deadlock the old exemption hid rather than solved. Pinned by `test_next_action_mutual_wake_only_producers_do_not_deadlock`.

**Blast radius**: the simplifier only. No other role in the default refine/plan/implement graphs carries a wake-only edge.

## Liveness

Unchanged in kind from #3478 — the gate holds only while an ungated *terminal* producer is pre-propose:

- lifts on the upstream's first proposal of any kind (current, withdrawn, or no-changes);
- dissolves entirely when the upstream is excused — `excuse_producer` removes every edge targeting it, wake-only included. This is newly load-bearing (before, the simplifier polled regardless and could not be stranded), so it now has a regression test;
- a crashed upstream resolves via supervision / HITL exactly as for the `tester` and `risk_analyst` gates.

Secondary effect, in the right direction: a `wait` role is exempt from `_check_convergence_stall`, so the simplifier also stops generating `stuck-phase-transition` anomalies that pointed at it while the upstream was the actual problem.

## What this does not do (#3603 remainder)

**Suggested fix 1 — "do not park a role whose heartbeat names an unsatisfied upstream producer in `waiting_on`."** Fixed at the derivation layer instead of the supervisor layer, deliberately. Wiring `waiting_on` heartbeat metadata into the *park decision* would invert `_role_waiting_status`'s fail-safe direction: it is documented (#3520) so that a stale or unreadable self-report can only make the alert **more** alarming, never quieter, because a genuinely wedged-but-polite agent emits the same `WAITING_ON_ROLE` heartbeat as a correctly-blocked one. As a spawn gate, a false or stale positive would suppress the park indefinitely and restore the ~50-pod respawn storm #3425 exists to stop. The tracker already holds the dependency deterministically, so the ordering fix needs no heuristic and costs zero pods.

That leaves a residual: a role correctly blocked on a dependency the review graph does **not** model can still park. In the default graphs there is no such role — every downstream-producer relationship is an edge, and all three are now gated. A custom graph that omits an edge would still be exposed; the fix there is to model the edge.

**Suggested fix 2 — "never park the sole remaining producer, or force an immediate probe plus escalation."**

- *Escalation already exists* and is what produced the log line quoted in the issue: `_check_arms_parked` (#3548) fires on the first poll where every derivable spawn arm is blocked with nothing in flight, and `_handle_arms_parked` raises an `event-arms-parked [high]` OVERSEER_ALERT **and persists a HITL decision** whose "Retry arms" option releases the parks in-band. That is ~5s after the wedge forms, not 1800s. Covered by `orchestrator/tests/test_arms_parked_stall.py`.
- *Not parking the sole producer* would reinstate the unbounded respawn loop in the case where pods are most expensive, and the issue itself notes respawning unchanged cannot converge.
- *A forced immediate probe* is not safely bounded by the existing latches: a probe spawn repopulates `_live_keys`, which clears the wedge condition, which resets the sticky latch, so the next no-op re-parks and re-triggers the forced probe — a 5s-cadence spawn loop. Bounding it needs a separate per-episode counter; that is a real design decision, not a drive-by, and the escalation above already gives the operator the signal within one poll.

Happy to split either into its own issue if you'd rather have them tracked separately.

## Testing

`make test` (full suite, 22,014 passed). 4 pre-existing local-environment failures, none reachable from this diff: two `tests/scripts/test_reap_stale_egg_images.py` cases (shell script exits 127 — missing container CLI in the sandbox) and two `WORKTREE_BASE_DIR` / `is_repos_parent_directory` path assertions in `gateway/tests/test_git_client.py` + `orchestrator/tests/test_per_slice_brc_commit.py`.

`make lint` clean.

New / revised tests in `orchestrator/tests/test_consensus_next_action.py`:

| Test | Pins |
|---|---|
| `test_next_action_wake_only_upstream_gates_first_propose` | the gate fires and lifts on the upstream's proposal (replaces `..._never_gates`) |
| `test_wake_only_simplifier_woken_to_propose_by_upstream_proposal` | the revised wake: repeated pre-proposal derivations stay `wait` (was: stay `propose`) |
| `test_next_action_wake_only_edge_still_casts_no_verdict` | #3381 not re-rolled — no spawn-able `ack`, no verdict |
| `test_next_action_mutual_wake_only_producers_do_not_deadlock` | acyclicity under mutual wake-only edges |
| `test_next_action_refine_phase_simplifier_gated_on_refiner` | the real refine graph — the 4/4 reproduced shape |
| `test_next_action_plan_phase_simplifier_gated_on_task_planner` | the real plan graph; `architect`'s proposal alone does not lift it |
| `test_next_action_wake_only_gate_dissolves_when_upstream_excused` | liveness under `excuse_producer` |

## Docs

`docs/architecture/orchestrator.md` (first-propose gate) and `docs/reference/agent-roles.md` (`simplifier`) both described the old poll-loop wake; updated.

Closes #3603


jwbron:
Follow-ups filed for the two parts of #3603 this PR deliberately leaves alone:

- #3608 — the unattended half of suggested fix 2 (forced probe on the all-arms-parked wedge, with the latch-loop analysis showing why it needs a bound independent of `_arms_parked_alerted`).
- #3610 — the residual of suggested fix 1: the park decision is still dependency-blind for dependencies the review graph does not model, plus why wiring `waiting_on` into the park decision was rejected.

"Closes #3603" still stands: the reproduced 4/4 case is fixed here, and neither follow-up is a regression introduced by this PR.

Issue #3616, opened 2026-07-25
Pin sampling parameters per model so provider defaults cannot drift silently


jwbron:
Split out of #3599 (item 2 of 3). Item 1 — recording the decoding config of every routed call — shipped in #3611; this is the remaining "neither set" half of that issue's title.

## Problem

egg pins **no** sampling parameters for any agent. The effective configuration is: Claude Code CLI defaults, then LiteLLM translation, then the provider's server-side defaults. None of it is pinned, and **it can change under us with no egg change** — a provider adjusting its default `temperature` or `top_p` silently changes agent behaviour across every pipeline, and nothing in egg would register that anything happened.

With #3611 merged this is now directly observable rather than inferred. On a stock route the per-call log line reads:

```json
"request_params": {"max_tokens": 32000, "stream": true, "extra_body": {"provider": {...}}}
```

No `temperature`, no `top_p`, no `top_k`, no penalty — each absent key means the parameter was never sent and the provider's own default applied.

## Why it matters

Repetition and degeneration are decoding-sensitive failure modes. #3598 documents a `laguna-s-2.1` producer emitting 480 identical tool calls out of 603; the leading candidate explanations are decoding-sensitive. Without a pin, two runs a month apart are not comparable, and any threshold or bisection measured under one configuration does not transfer to another.

## Suggested work

Set explicit `temperature` / `top_p` per model in `litellm_params`, checked against each provider's published recommended sampling for that model.

## Open questions to settle first

1. **Where does the pin live?** The model configs in play are operator-local (`~/.config/egg/litellm-models.yaml`); `laguna-s-2.1` appears nowhere in this repo. The repo carries only `config/litellm-models.template.yaml` with one example backend. Pinning per-model values into a template that operators copy once creates values that go stale silently — the same drift this issue exists to prevent. Deciding the mechanism (template defaults? a checked-in per-model recommended-sampling map? a validation step in `make litellm-config`?) is most of the work.
2. **It contradicts an existing deliberate choice.** `config/litellm-models.template.yaml` currently advises: *"Leave temperature / top_p / penalties at defaults for cleaner cross-provider comparisons."* That stance was chosen on purpose for the cq-6 diversification pilot and needs to be revisited explicitly, not overwritten.
[...]

Issue #3599, opened 2026-07-25
Sampling params are neither set nor recorded, making model-behaviour incidents unanalysable


jwbron:
## Summary

egg sets **no sampling parameters** for any agent, and **records none**, so when a model
misbehaves there is no way to determine what decoding configuration produced the behaviour —
during the incident or afterwards. This directly blocked the investigation in #3598, where a
`laguna-s-2.1` producer livelocked twice and the leading candidate explanations are
decoding-sensitive.

## Unset

Traced end to end; egg contributes exactly one generation-affecting parameter anywhere.

- `ClaudeAgentOptions` (`shared/egg_agent/client.py:363`) sets `permission_mode`, `model`, `cwd`,
  `env`, `setting_sources`, `disallowed_tools`, `can_use_tool`, `max_buffer_size`, plus optional
  `max_turns` / `system_prompt` / `effort`. **No `temperature`, `top_p`, `top_k`, or any penalty.**
- No per-model sampling params in the LiteLLM model config, and `litellm_settings` carries only
  the cost callback.
- Every model entry sets `drop_params: true`, so anything unsupported is discarded silently.

The only generation-affecting parameter egg sets is `reasoning_effort: high` in `litellm_params`.

So the effective configuration is: Claude Code CLI defaults, then LiteLLM translation, then the
provider's server-side defaults. None of that is pinned, and it can change under us without any
egg change.

## Unrecorded

`config/litellm/cost_callback.py` captures cost and cache statistics per call but **not
`optional_params`**, and no request-body logging is configured. Consequently the parameters for
any past run are unrecoverable once the pod is gone.

Repetition and degeneration are decoding-sensitive failure modes: temperature, top_p, and
repetition/frequency/presence penalties all materially change their probability. So the one class
of incident where this data matters most is precisely the class egg cannot reconstruct.

**This already had a concrete cost.** In #3598 a producer emitted 480 identical tool calls out of
603. Whether that is inherent to the model or an artifact of an unlucky decoding configuration is
unanswerable, because nothing recorded which configuration was in play. The issue remains open on
exactly that point.

## A structural note on penalties

Agents reach the upstream through LiteLLM's `anthropic_response` endpoint. The Anthropic Messages
API surface has **no repetition, frequency, or presence penalty field** — only `temperature`,
`top_p`, `top_k`. So the single most relevant anti-repetition control cannot be sent by the client
at all on this route; it would have to be injected server-side via `litellm_params` / `extra_body`,
[...]

PR #3620, opened 2026-07-25
fix(litellm): stop reasoning_effort being a silent no-op on OpenRouter routes

jwbron:
> **Stacked on #3611** (base is `egg/litellm-record-request-params`, not `main`). Review that one first; this PR's diff is the two commits' difference. Rebase to `main` once #3611 lands.

## The bug

egg's operator config sets `reasoning_effort: high` in `litellm_params`. Per #3599 that is **the only generation-affecting parameter egg sets anywhere**. It never reaches the wire for the models egg actually routes.

Found while verifying the unresolved question left open in #3599 and #3616. The answer turned out to be worse than the question: `reasoning_effort` isn't *overridden* by a client `thinking` budget, it is *dropped outright*, with or without one.

## Root cause

`OpenrouterConfig.get_supported_openai_params` advertises `reasoning_effort` / `thinking` only when `litellm.supports_reasoning(model)` is true, and that reads LiteLLM's built-in model-cost map. A model **absent** from the map answers `False`, so the gate fails closed — and `drop_params: true` then discards the parameter in a bare `pass`: no exception, no log line.

Absent is the *normal* state for an OpenRouter slug. Verified against the pinned litellm 1.86.2:

| Model | In LiteLLM's map | `supports_reasoning` |
|---|---|---|
| `qwen/qwen3-max` | ✗ | `False` |
| `moonshotai/kimi-k2-thinking` | ✗ | `False` |
| `deepseek/deepseek-r1` | ✓ | `True` |

OpenRouter adds models faster than LiteLLM's map tracks them, so a new slug is unflagged by default and the agent quietly runs at the provider's default reasoning depth. This is the default case, not the exotic one.

## Verification

Driven through a `Router` `model_list` entry on the `anthropic_messages` route (the shape and route the real proxy uses), HTTP mocked at the transport layer:

| Config form | Wire body |
|---|---|
| `litellm_params.reasoning_effort: high` (today) | *absent* ✗ |
| `+ model_info: {supports_reasoning: true}` | `reasoning_effort: high` ✓ |
| `extra_body: {reasoning: {effort: high}}` | `reasoning: {effort: high}` ✓ |
| `drop_params: false` | request **fails**: `UnsupportedParamsError` (HTTP 500) |

Controls run in the same harness: `temperature`, `top_p`, `seed` and the `extra_body` provider pin all reach the wire from `litellm_params` unchanged, so this is specific to `reasoning_effort` and not a general "config doesn't apply" problem.

`drop_params: false` is not a fix at all. Unsupported params do not pass through; they raise
`UnsupportedParamsError` with `status_code=500` (`litellm/utils.py:4082-4085`), so the route 500s
instead of quietly under-reasoning. That flag is also what stops Anthropic-only fields LiteLLM cannot translate (e.g. `thinking`) from 400ing tool-heavy turns, as the template documents. It trades a silent no-op for loud request failures.

## Changes

- **Template** documents the trap and both working forms, and steers the `qwen3-max` example to the `extra_body.reasoning` shape — which bypasses the param mapper entirely and so does not depend on the model map at all. That makes it the more durable of the two for OpenRouter routes.
- **CI guard**: an entry setting `litellm_params.reasoning_effort` without `model_info.supports_reasoning: true` now fails the build. The predicate is unit-tested in both directions, including that the `extra_body.reasoning` form is correctly *not* flagged — a guard that cannot fire is not a guard.
- **Second guard** pins `model_info` across paired bare/`[1m]` rows. It is a sibling of `litellm_params`, so the existing params-parity test never saw it; divergence would mean probe traffic and real traffic run at different reasoning depths.
- **Guide** documents the trap with the verification table.

## The limit of the CI guard, and what covers it

Operator overlays (`~/.config/egg/litellm-models.yaml`) are not in CI, and the models in play — `laguna-s-2.1` — live only there. So the guard protects the committed template and nothing else.

That is why every surface here points at the same runtime check instead: `request_params` from #3611 reports **post-mapping** params, i.e. what the wire carried. If `reasoning_effort` is not on that log line, it did not reach the model, whatever the config file says. The guard prevents the mistake in the repo; the log line catches it everywhere else.

## Upstream

The general defect — `drop_params` discarding caller-specified params in total silence — is fixed separately in the litellm fork (`jwbron/litellm`, branch `drop-params-visibility`): a deduped warning naming the dropped params, model and provider, with 5 tests. Not yet pushed or proposed upstream; say the word. That fix is *not* in the pinned 1.86.2 image, so it does not change egg's runtime here.

## Testing

- 8 tests pass in `tests/config/test_litellm_template.py` (3 new predicate tests + the template assertion + the new `model_info` parity guard).
- 185 pass in `tests/config/`.
- `make lint` clean.

### DIFF
--- a/config/litellm-models.template.yaml
+++ b/config/litellm-models.template.yaml
@@ -110,14 +110,25 @@
 # `UnsupportedParamsError` (HTTP 500) instead. It trades a silent no-op
 # for loud request failures.
 #
-# OFF OPENROUTER, `model_info` IS NOT THE FIX. `supports_reasoning` is
-# read only by `OpenrouterConfig`; other providers answer from a static
-# list (e.g. `TogetherAIConfig` subtracts from `OpenAIGPTConfig`'s base
-# list, which carries no `reasoning_effort` at all). Setting the flag on
-# a `together_ai/*` row changes nothing and the parameter is still
-# popped silently — so confirm the provider's chat config advertises
-# `reasoning_effort` before setting it. Most OpenAI-compatible ones do
-# not. The CI guard tracks this separately from the OpenRouter case.
+# THIS IS NOT AN OPENROUTER-ONLY TRAP, and `model_info` is not an
+# OpenRouter-only fix. Verified at litellm 1.86.2, nine other provider
+# configs gate a reasoning knob on the same `litellm.supports_reasoning`
+# call, fail-closed in the same way: `groq`, `xai`, `deepinfra`,
+# `cerebras`, `fireworks_ai`, `perplexity` and `bedrock_mantle` gate
+# `reasoning_effort`; `zai` and `minimax` gate `thinking` (and never
+# advertise `reasoning_effort` at all). On all of them form 1 works,
+# because LiteLLM registers `model_info` into its model-cost map under
+# the raw `litellm_params.model` string whatever the provider.
+#
+# Where `model_info` really is a placebo is a provider whose chat config
+# never consults `supports_reasoning` — e.g. `TogetherAIConfig`, which
+# only subtracts from `OpenAIGPTConfig`'s base list, and that base list
+# carries no reasoning knob at all. On a `together_ai/*` row the
+# parameter is popped silently whatever you set. So: check the
+# provider's `get_supported_openai_params` before setting the knob, and
+# note that the answer is per-knob, not per-provider. The CI guard
+# tracks the two cases separately (see the provider tables in
+# tests/config/test_litellm_template.py).
 #
 # VERIFY, don't assume: the `cost_callback` logs `request_params` per
 # call, read from LiteLLM's post-mapping params — i.e. what the wire
--- a/config/litellm-models.template.yaml
+++ b/config/litellm-models.template.yaml
@@ -50,14 +50,14 @@
 #
 # ## reasoning_effort is silently dropped unless the model is flagged
 #
-# `litellm_params.reasoning_effort` reaches the wire only when LiteLLM
-# believes the model is reasoning-capable.
+# On `openrouter/*` routes, `litellm_params.reasoning_effort` reaches
+# the wire only when LiteLLM believes the model is reasoning-capable.
 # `OpenrouterConfig.get_supported_openai_params` advertises
-# `reasoning_effort` / `thinking` only if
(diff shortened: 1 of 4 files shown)


jwbron:
Closing in favour of #3624, which fixes the root cause instead of documenting a workaround for it.

**Why this is superseded.** The diagnosis here is correct and was independently re-derived: `OpenrouterConfig.get_supported_openai_params` gates on `litellm.supports_reasoning`, that reads the bundled model-cost map, every OpenRouter slug egg routes is absent from it, so the gate fails closed and `drop_params` discards the knob in silence. That part stands.

What does not survive is the remedy. This PR's CI guard requires `model_info: {supports_reasoning: true}` beside any `reasoning_effort`, and its 652-line provider table hand-mirrors LiteLLM internals pinned to 1.86.2. #3624 removes the need for both: jwbron/litellm#8 reads OpenRouter's published per-model `supported_parameters` directly, so no operator assertion is needed and no table needs re-verifying on every image bump. The guard would also have demanded an operator assert `supports_reasoning: true` for `poolside/laguna-s-2.1`, which OpenRouter does not list `reasoning_effort` for at all.

**And the recommendation would have been a regression.** This PR steers the template to `extra_body: {reasoning: {effort: high}}`. Measured directly against OpenRouter (flat key, `max_tokens: 16000`, n=4, mean reasoning tokens):

| Model | no parameter | `reasoning_effort: high` | `reasoning_effort: low` |
|---|---|---|---|
| `moonshotai/kimi-k3` | 3130 | 340 | 49 |
| `z-ai/glm-5.2` | 1689 | 1090 | 788 |

On kimi-k3 the distributions do not overlap (lowest baseline sample 2240 vs highest `high` sample 1000). The knob caps reasoning below the model default rather than raising it, so adopting the shape this PR recommends would have cut reasoning roughly 9x on the primary model. `reasoning_effort` has now been removed from the operator overlay entirely.

**What was salvaged:** the description of the trap, and the runtime check (`request_params` from #3611 reports post-mapping params, so absence there means the knob never reached the model). Both carried into #3624 and the patch notes in `config/litellm/patch_litellm_cache.py`.

Thanks to the reviewer: the three rounds of NACKs on this were technically right each time, including the `gemini`/`vertex_ai` omission, which I verified against the pinned image.

Issue #3639, opened 2026-07-26
Convergence-stall respawn destroys uncommitted work: 110 minutes and 33 files lost from a healthy, productive coder


jwbron:
## Summary

The convergence-stall detector fires on **elapsed time since an actionable event was dispatched**,
with no reference to whether the agent is actually making progress. Its remedy is a respawn, and the
respawn's worktree re-attach runs a clean/sync that **discards all uncommitted work**.

Net effect: an agent doing genuine, continuous work loses everything at the 30-minute budget unless
it happens to have committed. Salvage does not help — it recovers commits, and there are none.

Observed on `issue-3632-v1`: **110 minutes of implementation across 33 modified files destroyed**,
silently, on a coder that was demonstrably healthy and productive throughout.

This is the exact scenario #3596 was opened to prevent — an operator unable to distinguish an agent
doing real work from one whose work is "silently discarded every cycle". Here the harness is the
thing discarding it.

## Timeline

The coder was spawned for a `propose` event and worked continuously. Measured from its live session
transcript at the 109-minute mark, minutes before it was killed:

```
596 tool calls, 587 distinct, 30 novel inputs in the last 30 calls
tools: Read 207, Bash 169, Edit 113, TaskUpdate 11, TaskCreate 8
git status: 33 modified files
git log origin/egg/issue-3632-v1/work..HEAD: 0 commits
```

Modified files were exactly the change surface the plan called for — `peer_consensus/__init__.py`,
`redis_message_store.py`, `concurrent_executor.py`, `routes/consensus.py`,
`routes/decisions/_graph_mutations.py`, `routes/phases/_advance.py`, and more. Its last actions were
editing tests and running `pytest`. This was not a stuck agent by any measure available.

Then:

```
03:20:13  Convergence-stall detected: role=coder actionable event 'propose' pending 3610s
          (budget 30 min=1800s); raising stuck-phase-transition anomaly     [event_loop/_loop.py:914]
03:20:18  Worktree re-attach: validation succeeded (cleanup pending)  agent_worktree_id=issue-3632-v1-coder
03:20:19  Worktree re-attach: cleaned and synced                      agent_worktree_id=issue-3632-v1-coder
03:20:19  Agent Job created job_name=egg-agent-issue-3632-v1-coder-ace18ef0  action=propose
```

The replacement pod, 15 seconds old:

```
commits ahead: 0
dirty files:   0
[...]

Issue #3509, opened 2026-07-06
Dirty-exit discard is unrecoverable when agent memory dies: auto-salvage orphans and record discarded tips durably


jwbron:
## Incident (pipeline issue-3312-v2, slice-4, 2026-07-06, build eca7ca740; third dirty-exit discard)

Follow-up to #3506. The first two dirty-exit discards self-healed instantly (the coder fast-forward merged back to its recorded tip). The third one hit a compounding failure that made the orphaned work unrecoverable BY THE AGENT, causing a silent multi-day regression:

1. R6 dirty-path discard threw away the coder's 52-commit stack (tip e953a9525, which included the completed giant-function splits: `_run_pipeline` under cap at 1,467 lines, i.e. the program's non-negotiable #7 essentially done).
2. Both of the coder's memory sources died simultaneously:
   - Redis session state had passed its 6-hour TTL (overnight gap), so no session resume.
   - Its durable BRC memory commits lived IN the discarded stack (memory rides the same branch lineage it describes), so the discard erased the map to itself.
3. Result: the coder woke amnesiac, re-anchored to the oldest orphan it could find (0228f4a9f, the original baseline from two days prior), and silently began RE-DERIVING the entire decomposition. Pipeline status showed "running, all agents WORKING" throughout; no failure, no decision, no alert. The regression was only discovered by host-side reflog inspection.

The discard warning added by 3507 helpfully says "salvageable via salvage_agent_commits (#3368)", but nothing invokes it, the discarded tip SHA is recorded nowhere durable, and `salvage_agent_commits` inspects worktree branches, which at that point already pointed at the re-derivation lineage; the orphan existed only in the object store.

## Asks

- **Auto-salvage on discard**: when the R6 path discards commits, immediately push the discarded tip to `egg/recovered/<pipeline>/<slice>-<role>/<sha>` through the ledger-aware path (same mechanism as #3368's salvage), not just log a warning. Objects in the store are one gc away from gone and invisible to everyone meanwhile.
- **Record the discarded tip durably where the agent looks**: contract note, message-bus system message to the role, or a branch-note; so a resuming agent (even one with zero session memory) deterministically finds "your prior tip was X, a fast-forward descendant of base" instead of depending on expired TTLs or memory commits that rode the discarded lineage.
[...]

Issue #3506, opened 2026-07-05
Re-attach hard-sync discards unpushed agent commits between event-pump cycles; multi-session slices cannot accumulate work


jwbron:
## Symptom (pipeline issue-3312-v2, slice-4, 2026-07-04 23:00Z onward, build f348ed166)

The slice-4 coder has run ~6 productive sessions over 3+ hours (one every ~30 min, up to 10 min each, ~$1/session) and the slice branch has ZERO net progress: `egg/issue-3312-v2/slice-4`'s tip is still the chain base (64fa30773, identical to slice-3's tip).

## Root cause

Worktree re-attach's R6 dirty-state handling hard-resets the agent worktree to the remote branch tip on EVERY event-pump cycle. The coder's worktree reflog is an unbroken chain of:

```
64fa30773 HEAD@{0}: reset: moving to origin/egg/issue-3312-v2/slice-4
64fa30773 HEAD@{1}: reset: moving to HEAD
64fa30773 HEAD@{2}: reset: moving to origin/egg/issue-3312-v2/slice-4
...
```

Any commit an agent makes locally but does not push within the SAME cycle is discarded by the next cycle's reset. Concrete casualty: the coder's slice-4 baseline commit

```
0228f4a9f "Move orchestrator/routes/pipelines.py to pipelines/__init__.py (pure move baseline)"
parent 64fa30773 (exactly the branch tip; clean fast-forward), author egg, committed 23:42Z
```

now orphaned in the object store. It was created in the session that ended with the coder reporting "Branch state is clean and the baseline is safely committed"; the next re-attach wiped it. The coder's durable session memory still believes the baseline exists, so subsequent sessions operate on a false premise; a Sisyphus loop that burns cost with zero durable progress and no error signal anywhere (gateway shows no push denials; the coder simply did not push, and nothing requires it to).

## Why this never bit before

Slices 1-3 of this pipeline (and prior programs) did each slice's coding in a SINGLE long session that ended with a push at propose time, so the reset-to-remote between cycles was a no-op. Slice-4 (orchestrator/routes/pipelines.py, 30.5k lines, 3 functions individually over 1,500 lines) is the first slice whose coder plans multi-session work before proposing. The architecture makes multi-session local accumulation structurally impossible.

## Asks

- **Make re-attach fast-forward-aware**: if the worktree's local HEAD on the assigned branch is a strict descendant of the remote tip (clean tree), keep it instead of resetting. Reset only on divergence or dirt.
[...]

Issue #2807, opened 2026-05-27
Coder loses uncommitted Edits when agent crashes mid-task


jwbron:
## Background

Surfaced during the #2804 investigation. Distinct enough to track
separately.

When an agent (typically the coder) crashes between landing a
``Write``/``Edit``/``NotebookEdit`` to disk and calling ``git commit``,
**the on-disk changes are lost** when the wrapper respawns a fresh
container. The next container starts from the branch HEAD, so any
filesystem state from the previous attempt — partial implementation,
half-applied edits, in-progress refactor — silently disappears.

## Observed on `issue-2777`

From the #2804 16:30:47 UTC crash trail:

- Coder had just applied an ``Edit`` to ``phases.py:500`` wiring
  ``_open_context_pr_at_implement_start`` into the plan→implement
  boundary (the exact wiring the prior reviewer NACK was requesting).
- Coder was mid-second-``Edit`` (wiring
  ``validate_plan_for_implement_phase``) when the SDK 1MB buffer
  overflow killed the agent process.
- Neither edit was committed. Container respawned, branch HEAD
  unchanged, both fixes need to be redone from scratch.

With 11 of 14 coder tasks still TODO at crash time, this single
crash wiped ~10 minutes of validated LLM work — the coder had
already addressed reviewer feedback locally, but the artifact never
reached a branch where the next container could pick up.

## Why this matters

The crash recovery model assumes the agent's persistent state is the
git branch — fresh container, ``git checkout``, resume work. That
assumption breaks whenever the agent's *most recent valuable work*
lives in unstaged or uncommitted form, which is essentially the
window between ``Edit`` and the next ``git commit``.

Today the coder accumulates 3-4 task edits before committing
(commit at consensus-propose time, not per-task). That's exactly
the window a crash steals work from. Less frequent crashes would
reduce the cost; #2804 attacks the crash rate. But work loss is a
real failure mode on its own — OOM, network blips, SDK regressions,
or any other agent-process death exposes it.

### Option A — Aggressive ``git commit`` directive (cheap)

Add a coder system prompt directive: ``commit after each task
completion, even if the artifact isn't ready to propose``. Trivial
to ship as a sandbox prompt update.

**Tradeoff:** producer-role agents commit transparently throughout
their work, then re-propose from the latest commit. Reviewers
already see only committed artifacts via the BRC ``read_peer_artifact``
flow, so per-task commits don't change reviewer behavior. The risk
is committing broken intermediate state — but the coder runs ``make
[...]

PR #3644, opened 2026-07-26
Preserve a re-attached worktree's uncommitted work before the R6 reset (#3639)

jwbron:
Fixes the work-destroying half of #3639.

## What was actually destroying the work

`_clean_reused_worktree` hard-resets a reused worktree on every event respawn. #3506/#3509 built preservation around that reset, but both operate on **commits**: `salvage_discarded_tip` pushes the doomed HEAD to `egg/recovered/...` only when the orphan detector finds commits ahead of the origin tip. A session that worked for hours without committing gave it nothing to find, so `git reset --hard` + `git clean -fd` erased the working tree and the path logged `cleaned and synced` at INFO. That is the #3639 incident: 110 minutes across 33 modified files, zero commits, no recovery ref.

## The change

Close the gap one step earlier instead of building new machinery. When the pre-discard tree is dirty, `_preserve_dirty_tree` commits it (`git add -A` plus a `[salvage] pre-reset working-tree state (#3639)` commit) **before** the reset. The snapshot is then an ordinary orphan, so the existing #3509 salvage push and message-bus record carry it to a recovery ref with no new subsystem.

The residue policy is unchanged. `was_dirty` is latched before the snapshot, so the tree still hard-resets to the origin tip and a successor never inherits a killed-mid-event working set; the snapshot only makes the discarded state recoverable.

Details:

- **Best-effort throughout.** A failed add/commit, or a tree holding only ignored files, logs at WARNING with the entry count and proceeds with the reset. Blocking reuse would only divert the spawn to create-with-retry, which destroys the same state with less visibility.
- **Skipped when `branch` is None.** With no origin tip and no salvage target, a snapshot would just become the successor's HEAD: un-vetted residue promoted to committed state, exactly what R6 exists to prevent. That case now logs the discard rather than staying silent (issue suggestion 4).
- **Reuses #2807's `egg-salvage` identity** so one `[salvage]` grep finds every machine-made working-tree snapshot regardless of which path took it. It does not call `agent_salvage.commit_working_tree` directly: that helper's `_run_git` omits `safe.directory=*`, which the re-attach path needs for host-uid-owned worktrees, so reusing it would fail on "dubious ownership" exactly in production. The caller's git closure is threaded in instead.
- **The bus message** tells a resuming agent the top commit is an automatic snapshot to review, not work it already proposed, and carries `wip_commit` in its metadata.

## Correction to the issue's diagnosis (does not change this fix)

The issue attributes the kill to the convergence-stall detector. It is not the trigger, and I posted the full evidence on #3639. In short, from the orchestrator log of the incident:

- `_check_convergence_stall` has no remedy. Its only effect is `_emit_supervision_alert`, an informational `OVERSEER_ALERT` bus message (`classify_message_intent` explicitly renders `stuck-phase-transition` as non-binding). It kills nothing and respawns nothing.
- Both agent Jobs were already **terminal** when the orchestrator touched them at 03:20:13: they were deleted with `propagation=Background`, which is only reachable from `reap_terminated`, and `_reap(only_terminal=True)` skips any Job that is not FAILED/EXITED.
- Coder spawned 01:19:58, documenter 01:20:00; both terminal at 03:20:13. That is the 7200s (2 hour) agent execution timeout in `shared/egg_agent/client.py`, which returns `returncode=-1` and fails the Job.

So the chain was: agent hit its own 2h timeout -> loop observed a terminal Job and respawned (correct behavior) -> the respawn's re-attach wiped the tree (the defect this PR fixes). The stall warning fired in the same poll tick because the dedupe key had just left `_live_keys` and the BRC bus had been quiet 3610s. Making the stall criterion progress-aware would have prevented a misleading alert, not the loss.

Left for follow-ups, since both rest on different premises than the fix above: the progress-blind stall alert (#3639 suggestions 1 and 3), and the 2h agent timeout that kills a productive session with no warning and no checkpoint.

## Tests

`orchestrator/tests/test_kubernetes_spawner.py::TestDirtyTreePreservedBeforeReset` (new, 6 cases): dirt with zero commits is snapshotted, pushed to the recovery ref, and reported on the bus while the successor still starts clean at the origin tip; ignored-files-only dirt is not committed; a failed snapshot still resets; the helper swallows git failures; a clean tree is untouched. The two central cases fail against `main`.

Existing `TestDirtyDiscardAutoSalvage` cases that seeded dirt alongside a commit were updated: the salvaged tip is now the snapshot sitting on the predecessor's commit (count 2, not 1), asserted structurally via parentage rather than a hardcoded sha.

`make lint` clean; the file crosses the 800-line soft cap (881), which is a non-fatal warning and well under the 1500 hard cap.

Docs: the re-attach policy in `on-demand-agent-lifecycle.md` and the recovery reference in `agent-recovery.md` previously stated only that uncommitted state is discarded. Per the issue's second comment ("an agent cannot follow a contract it is never shown"), both now record the snapshot behavior.


jwbron:
Follow-ups from the analysis in this PR's description are now filed: #3658 (the 7200s agent session timeout that actually ended the session) and #3659 (the progress-blind stall alert, scoped as an alerting fix). Neither blocks this change.

Issue #3692, opened 2026-07-28
Make egg tolerant of weak-termination models: force the terminal action, suppress degenerate repeats, and check correspondence deterministically


jwbron:
> **Goal: make egg able to run cheap models unattended.** Six runs of the `laguna-s-2.1` assessment
> (#3595) established that this class of model produces competitive *output* and cannot reliably
> *terminate*. The remedy belongs in egg, not in the model. This issue is the work.

## Why

From #3595, across six pipelines:

| Measure | Result |
|---|---|
| Livelocks | **20** (19 through the run-6 tally, plus the live-observed run-6 tester relapse), five roles, all three phases. Renumbered 2026-07-29; see the numbering correction on #3595 |
| Self-recovered | **0 of 18** |
| Detected by the product | **0 of 18** |
| Cleared by a formulaic operator message | **17 of 18** (the 18th was ended by a blunt 7200s timeout) |
| Operator interventions, run 6 | **12** in ~7.5 h — one per ~37 min |
| Pipelines completed end-to-end, any run | **0** |

Every livelock has the same shape: **the agent announces the terminal action, then issues a read
instead, indefinitely.** It is never confused about the task. It fails to *emit* the call. Run 1's
refiner said *"I have all the information I need"* 119 times; run 6's refiner said *"Let me now
write my refine analysis"* and re-read a file 40 times.

That shape is attackable structurally. **Forcing beats detecting**: the fix is to make the loop
impossible rather than to observe it and correct it afterwards.

> **The mechanism now says this is not just the best of several options, it is the only one that
> works** ([analysis](https://github.com/jwbron/egg/issues/3693#issuecomment-5110977007)). 36 of
> the model's 48 layers are sliding-window at **512 tokens** and the livelock's repeat cycle was
> **243 tokens**, so three quarters of the network sees ~2 verbatim copies of one cycle and
> nothing else. That distribution is near-degenerate. **No message, at any length or phrasing,
> injected into that context can outbid it** — which is precisely what the four inert mid-turn
> nudges in run 1 measured. Only removing the copies (`fresh_session`) or removing the alternative
> action (slice 1) changes the outcome. It also bounds the cost of slice 1: the attractor is
> established within **two** iterations, so detection latency is not what sets `K`.

## Not in scope here — already in flight

Do not duplicate: **#3686** (→#3681 propose-gate docs/`full_command`), **#3687** (→#3658 session
[...]

PR #3698, opened 2026-07-29
litellm: carry prior-turn assistant reasoning to OpenRouter (Patch 10)

jwbron:
> Correction after merge: the "keeps the seed alive" framing below is too strong. See
> https://github.com/jwbron/egg/pull/3698#issuecomment-5122141626

## The defect

On egg's primary route (Claude Code -> gateway -> LiteLLM `/v1/messages` -> OpenRouter),
**assistant reasoning from prior turns was never sent back to the model.**

Verified against pristine `BerriAI/litellm` at v1.86.2, the version the image pins:

1. LiteLLM's Anthropic adapter converts incoming `thinking` content blocks into
   `assistant_message["thinking_blocks"]`
   (`llms/anthropic/experimental_pass_through/adapters/transformation.py:662-663`).
2. `llms/openrouter/chat/transformation.py` has **no request-path consumer** for that field. Its
   only reasoning line is response-path (`reasoning` -> `reasoning_content` on streaming deltas).
3. `reasoning_details` appears nowhere under `llms/` or `litellm_core_utils/`.

`OpenAIGPTConfig.transform_request` puts `messages` straight into the request body, so
`thinking_blocks` reached OpenRouter as a field no one reads and every historical assistant turn
arrived with its reasoning missing.

For a model whose chat template re-renders prior thinking this is a malformed history rather than
a lost optimisation. Poolside Laguna renders
`'<think>' + message.reasoning|message.reasoning_content + '</think>'` for every previous
assistant turn, so each one became a literal empty `<think></think>`; Poolside's model card warns
this degrades follow-up behaviour.

**The gap is upstream's.** It predates every egg and `jwbron/litellm` change. The fork's OpenRouter
work (jwbron/litellm#8) is confined to `get_supported_openai_params`, which acts on
`optional_params` and cannot reach a message field.

## Precondition check

The fix is only worth making if Claude Code actually returns thinking blocks in the `/v1/messages`
request body for a non-Anthropic model, so that was confirmed first, by capturing real request
bodies through a logging proxy in front of the LiteLLM proxy.

It does, under egg's configuration. A two-turn tool-calling exchange against
`deepseek/deepseek-v4-pro` produced a follow-up request whose assistant message carried
`['text', 'thinking', 'tool_use']`, with the block shaped
`{"type": "thinking", "thinking": "...", "signature": ""}`.

Two findings worth recording:

- **On `poolside/laguna-s-2.1` the blocks do not come back, and the reason is not what this PR
  originally said.** An earlier version of this body said that route "returned no reasoning content
  to begin with". The outcome was right and the cause was wrong, and the difference decides whether
  this patch is worth landing. See the section below.
- **The host `cllm` launcher is not representative of egg.** It sets
  `ANTHROPIC_CUSTOM_MODEL_OPTION_SUPPORTED_CAPABILITIES=""`, which egg deliberately does not set
  (`docs/guides/per-agent-models.md:416-425`). The precondition was therefore re-run under
  egg-equivalent environment before being accepted.

## Laguna: the patch is load-bearing, and it still cannot fire today

This section supersedes the original claim that Laguna "returns no reasoning at all". It does
reason. What it does not do, on egg's traffic, is ever *start*.

### The patch does real work, and the control is what proves it

Three arms, identical history, `max_tokens` 800, tools present in every arm. The only thing that
moves is what the prior assistant message carries. Full message sequences are in the reproduction
record on #3595.

| arm | prior assistant message carries | turns that reasoned | median `reasoning_tokens` |
|---|---|---|---|
| pre-patch | nothing | **0/4** | 0 |
| post-patch | `reasoning_content: <prior reasoning>` | **4/4** | 105 |
| control | the same text as plain `content` | **0/4** | 0 |

The control is the load-bearing cell. Identical text, identical token count, attached to a
different field: 4/4 against 0/4. It is specifically `reasoning_content`, which is exactly what this
patch produces, and not extra context, that restores the channel. The same result holds when the
next turn also emits a tool call (0/3 against 3/3, both arms calling the tool in every draw), so it
sustains through an agent loop rather than only on final answer turns.

Traced end to end through a real patched v1.86.2: a reasoning-bearing response comes out of the
adapter as `['thinking', 'text', 'tool_use']`, so Claude Code echoes a thinking block back, and this
patch maps it to `reasoning_content` on the next request. A response with no reasoning channel comes
out as `['text', 'tool_use']`, and there is nothing to map.

### Why it is nonetheless inert on egg's Laguna traffic right now

Because nothing seeds the channel. A tool-bearing opening turn on an errand-shaped prompt does not
reason, so there is no first reasoning to carry forward, and a patch that replays reasoning has
nothing to replay. Four-turn sessions, two draws each:

| arm | opener | replay | `reasoning_tokens` per turn |
|---|---|---|---|
| A | errand, tools | no | `[0,0,0,0]` `[0,0,0,0]` |
| B | errand, tools | **yes** | `[0,0,0,0]` `[0,0,0,0]` |
| C | **tool-free** | **yes** | `[1601,746,309,70]` `[1448,490,368,237]` |
| D | tool-free | no | `[1198,0,0,0]` `[1720,0,0,0]` |

Arm B is this patch alone: worthless without a seed. Arm D is a seed alone: it fires once and is
gone by the very next turn. Arm C is both, and it is the only arm that works. Arm D is also exactly
the shape production shows, a session total of reasoning tokens against otherwise uniform per-call
zeros.

So the original "inert on this route" was right about the outcome and wrong about the cause. Not
"the model does not reason", but "nothing ever primes it to".

### The seed is now known, and it is a per-request parameter

Run round-robin with cells interleaved, because the unprimed rate drifts over time and blocking the
cells would confound drift with cell. n=4 per cell, `max_tokens` 2500, fresh opening turn.

| opening turn | reasoned | `reasoning_tokens` |
|---|---|---|
| tools callable + errand prompt | **0/4** | `[0,0,0,0]` |
| tools + `tool_choice: "none"` + errand | 3/4 | `[236,0,485,1135]` |
| tools callable + reasoning-shaped | 3/4 | `[0,527,696,522]` |
| tools + `tool_choice: "none"` + reasoning-shaped | **4/4** | `[1334,1584,1261,2105]` |
| no tools + errand | 4/4 | `[246,638,821,105]` |
| no tools + reasoning-shaped | 4/4 | `[1791,1646,1311,1647]` |

Callable tools are the strong lever and `tool_choice` defeats them: 0/4 to 3/4 on a byte-identical
errand prompt with only `tool_choice` moved. Prompt shape is the weaker lever and mostly drives
magnitude.

**Recipe: open every agent session with one turn that sets `tool_choice: "none"` and is
reasoning-shaped.** 4/4 at 1261-2105 seed tokens, indistinguishable from dropping the tools array,
and it is a per-request parameter rather than a change to how egg assembles the tools array, which
makes it far cheaper to ship. That recipe is not in this PR; this PR is the half that keeps the seed
alive once it exists.

### Acceptance signal, per turn type, untested

- On **follow-up turns after a tool result**, per-call `reasoning_tokens` should be non-zero instead
  of uniformly zero.
- On **turns that emit a tool call**, the same, which is the stronger claim and the one that decides
  whether reasoning survives an agent loop at all.
- Session-wide averages are the wrong instrument here and should not be used: the pre-patch pathology
  is a per-turn zero, not a lower mean.

Nothing in this PR measures any of that on a real pipeline. It needs a run on the patched image with
the seeding recipe applied.

### What is untested, explicitly

- **Whether the chain survives real session lengths.** Arm C decays monotonically in both draws
  (1601, 746, 309, 70), which extrapolates to exhaustion somewhere around turn 6 to 8. egg's sessions
  run hundreds of calls. Four turns is not evidence about forty.
- **Whether a dropped chain can be revived.** Each turn is primed only by the previous turn's
  reasoning, so a single turn that happens not to reason may end the chain permanently. Untested, and
  it is the difference between a fix and a fix that needs re-seeding.
- The per-turn rate figure and the re-seed-on-dropout design note that circulated earlier are
  inference, not measurement, and are marked as such on #3692 and #3595.
- All of the above is 24 draws against one model in one afternoon, on a cell I have watched swing
  from 0/4 to 4/4 across time windows. Treat the recipe as a strong candidate, not a settled fix.

## The change

Patch 10 maps assistant `thinking_blocks` onto `reasoning_content` in
`OpenrouterConfig.transform_request` and removes `thinking_blocks` so no unknown field is
transmitted.

OpenRouter accepts `reasoning`, `reasoning_content` and `reasoning_details` interchangeably on an
assistant message and documents this for exactly this multi-turn tool-calling case; Poolside's
template reads `message.reasoning` / `message.reasoning_content`. The plain string form is used
because `reasoning_details` exists to carry encrypted or summarised blocks and the adapter
produces neither.

The logic lives in a **fourth staged module** (`config/litellm/openrouter_reasoning_roundtrip.py`
-> `llms/openrouter/_egg_reasoning_roundtrip.py`), following the same convention as Modules 1-3,
so it stays lintable and unit-testable in this repo rather than living as a string literal. Module
labels renumbered `N/3` -> `N/4` and patch labels `N/9` -> `N/10`; the Dockerfile stages the new
file; the module docstring gains the tenth entry.

Edge cases, all covered by tests: multiple blocks concatenate in order; `redacted_thinking` carries
opaque `data` and contributes nothing; a whitespace-only or empty result emits no field rather than
re-creating the empty `<think></think>`; `signature` is not forwarded; assistant messages only;
input is not mutated; fail soft, so a block that cannot be parsed leaves its message exactly as it
arrived; idempotent.

**Needle anchoring.** `_supports_cache_control_in_content` and `_move_cache_control_to_content` are
each named twice in 1.86.2, so neither call alone is unique. The needle spans the cache_control
pair *and* the following `extra_body` pop, a sequence that occurs only in `transform_request`.
Confirmed to match real stock v1.86.2 exactly once, with a test that pins the anchoring against a
sibling-call-site fixture (same discipline as Patches 4 and 8).

**Direction.** This is request-path (client -> provider). Patches 4, 5a and 5b are response-path
(provider -> client). Adjacent, not the same thing; a test asserts this patch never touches
`transform_response`.

## Validation

Applied the full script to a real stock `litellm==1.86.2` install: all 10 patches and 4 modules
applied, and a second run reported all 15 already applied (idempotent).

Replayed a real captured Claude Code `/v1/messages` body through that patched proxy with the
provider endpoint captured:

| | outgoing assistant message |
|---|---|
| before | `['role', 'thinking_blocks', 'tool_calls']`, no reasoning field |
| after | `['reasoning_content', 'role', 'tool_calls']`, no `thinking_blocks` |

Live two-turn tool-calling exchange against `deepseek/deepseek-v4-pro` through the patched proxy
completed normally. Per-turn growth of the assistant history on the follow-up turn: **118 tokens
before, 146 after (+28, 1.24x)** — the restored reasoning, exactly. The magnitude scales with how
much the model reasons; these probe tasks were small, so this is a floor rather than a typical
figure.

Note on a measurement that looks contradictory: total prompt tokens for the whole replayed body
went *down* (1992 -> 1982). That is Patch 3's billing-header filter also being active in the
patched build, not Patch 10; the isolated figure is the assistant-turn delta above.

`make lint` passes. `make test` has 4 failures, all pre-existing: verified by re-running them at
pristine `HEAD` with these changes set aside, where 3 fail identically
(`test_reap_stale_egg_images.py::TestReapScriptSafetyGuard` x2,
`test_git_client.py::test_worktrees_parent_detected`) and the 4th
(`test_error_paths.py::test_session_expires_exactly_at_boundary`) passes, i.e. a timing flake.
None are in `tests/config/`. The 114 tests in `tests/config/test_litellm_runtime_modules.py` and
`tests/config/test_patch_litellm_cache.py` pass.

## Fork PR

jwbron/litellm#11 implements the same change as ordinary source (`llms/openrouter/reasoning.py`
plus the `transform_request` call), with 27 tests in the fork's own layout; its full
`tests/test_litellm/llms/openrouter/` suite passes (154 tests). The two implementations are
AST-identical modulo docstrings and comments, checked mechanically, since the patch literal here is
what actually runs in production and divergence would be a bug. This PR does **not** make egg's
image build from the fork. A separate upstream `BerriAI/litellm` PR is a follow-up.

### DIFF
--- a/config/litellm/Dockerfile
+++ b/config/litellm/Dockerfile
@@ -5,7 +5,8 @@
 # stock image's Anthropic->OpenAI translation drops prompt-cache hits on
 # Qwen/DeepSeek, mis-streams reasoning models, silently drops params it does
 # not recognise, and manufactures a reasoning ceiling from the caller's
-# thinking budget. patch_litellm_cache.py closes the nine gaps at build time
+# thinking budget, and never sends prior-turn reasoning back.
+# patch_litellm_cache.py closes the ten gaps at build time
 # (see that script for the diagnosis);
 # cost_callback.py surfaces the resulting cost/cache stats into the pod log
 # stream. Both mirror the host-side dotfiles setup that took Qwen cache hit
@@ -31,6 +32,7 @@ COPY patch_litellm_cache.py /egg/patch_litellm_cache.py
 COPY openrouter_capabilities.py /egg/openrouter_capabilities.py
 COPY drop_params_visibility.py /egg/drop_params_visibility.py
 COPY anthropic_thinking_policy.py /egg/anthropic_thinking_policy.py
+COPY openrouter_reasoning_roundtrip.py /egg/openrouter_reasoning_roundtrip.py
 RUN python3 /egg/patch_litellm_cache.py
 
 # Custom cost/cache logger, registered as `cost_callback.cost_logger` under
--- a/config/litellm/openrouter_reasoning_roundtrip.py
+++ b/config/litellm/openrouter_reasoning_roundtrip.py
@@ -1,0 +1,125 @@
+"""Carry prior-turn assistant reasoning back to OpenRouter.
+
+egg's primary route is ``/v1/messages``: Claude Code -> egg-gateway -> LiteLLM
+-> OpenRouter. On the way in, LiteLLM's Anthropic adapter turns each assistant
+``thinking`` content block into an entry on
+``assistant_message["thinking_blocks"]`` (``_translate_anthropic_messages_to_openai``
+in ``llms/anthropic/experimental_pass_through/adapters/transformation.py``).
+Nothing on the OpenRouter **request** path then reads that field: stock
+``llms/openrouter/chat/transformation.py`` mentions reasoning exactly once, on
+the *response* side (``reasoning`` -> ``reasoning_content`` on streaming
+deltas), and ``reasoning_details`` appears nowhere in ``llms/`` or
+``litellm_core_utils/`` at all. So ``thinking_blocks`` rides to the provider as
+a field no one reads, and every historical assistant turn arrives carrying no
+reasoning.
+
+For a model whose chat template re-renders prior thinking that is not merely a
+lost optimisation, it is a malformed history. Poolside Laguna renders
+``'<think>' + message.reasoning|message.reasoning_content + '</think>'`` for
+every previous assistant turn, so each one arrives as a literal empty
+``<think></think>``; Poolside's model card warns that this degrades follow-up
+behaviour. The measured cost in egg was a livelock whose per-turn prompt growth
+was exactly +243 tokens, leaving no budget for reasoning content.
+
+This module maps the blocks the adapter produced onto the field OpenRouter
+actually reads. OpenRouter accepts ``reasoning``, ``reasoning_content`` and
+``reasoning_details`` interchangeably on an assistant message and documents
+this for exactly this multi-turn tool-calling case; Poolside's template reads
+``message.reasoning`` / ``message.reasoning_content``. We send the plain string
+form, which satisfies both: ``reasoning_details`` exists to carry encrypted or
+summarised blocks, and we have neither. Anthropic ``signature`` values are
+meaningless to OpenRouter and are not forwarded.
+
+The gap is upstream's, not egg's: it predates every egg and ``jwbron/litellm``
+change, and the fork's OpenRouter work (jwbron/litellm#8) is confined to
+``get_supported_openai_params``, which acts on ``optional_params`` and cannot
+reach a message field.
+
+Note the direction. Patches 4, 5a and 5b are response-path (provider ->
+client): they fix how reasoning is *streamed back*. This is request-path
+(client -> provider). They are adjacent, not the same thing.
+"""
+
+from typing import Any
+
+# Blocks the adapter can put on ``thinking_blocks``. ``redacted_thinking``
+# carries an opaque ``data`` payload rather than text (see
+# ``ChatCompletionRedactedThinkingBlock``); there is no plaintext to hand a
+# provider that wants a string, so those blocks contribute nothing.
+_THINKING_BLOCK_TYPE = "thinking"
+
+# The field the adapter writes and the field OpenRouter reads.
+_SOURCE_FIELD = "thinking_blocks"
+_TARGET_FIELD = "reasoning_content"
+
+
+def _extract_reasoning_text(blocks: Any) -> str:
+    """Concatenate the plaintext of ``blocks``, in order.
+
+    Non-dict entries, entries of any type other than ``thinking``, and entries
+    whose ``thinking`` is not a string are skipped individually rather than
+    failing the whole message: one malformed block should not cost us the
+    reasoning from its well-formed siblings. A ``blocks`` that is not a list at
+    all is a shape we do not understand, and the caller leaves the message
+    untouched.
+    """
+    if not isinstance(blocks, list):
+        raise TypeError(f"{_SOURCE_FIELD} is {type(blocks).__name__}, expected list")
+    parts: list[str] = []
+    for block in blocks:
+        if not isinstance(block, dict):
+            continue
+        if block.get("type") != _THINKING_BLOCK_TYPE:
+            continue
+        text = block.get(_THINKING_BLOCK_TYPE)
+        if isinstance(text, str):
+            parts.append(text)
+    return "".join(parts)
+
+
+def map_thinking_blocks_to_reasoning_content(messages: Any) -> Any:
+    """Return ``messages`` with assistant reasoning moved to the wire field.
+
+    For each assistant message carrying ``thinking_blocks``: drop that field so
+    no unknown key is transmitted, and set ``reasoning_content`` to the
+    concatenated plaintext when there is any. A message whose blocks yield only
+    whitespace (or nothing at all, e.g. ``redacted_thinking`` only) loses the
+    field and gains nothing, which is what OpenRouter would have received
+    anyway minus the noise.
+
+    User, system and tool messages are returned untouched, as is any assistant
+    message with no ``thinking_blocks`` key. Input is not mutated: touched
+    messages are shallow-copied.
+
+    Never raises. A message this function cannot make sense of is passed
+    through exactly as it arrived, so the worst case is stock behaviour.
+    """
+    if not isinstance(messages, list):
+        return messages
+
+    out: list[Any] = []
+    for message in messages:
+        try:
+            if not isinstance(message, dict):
+                out.append(message)
+                continue
+            if message.get("role") != "assistant" or _SOURCE_FIELD not in message:
+                out.append(message)
+                continue
+
+            text = _extract_reasoning_text(message.get(_SOURCE_FIELD))
+
+            updated: dict[str, Any] = dict(message)
+            updated.pop(_SOURCE_FIELD, None)
+            # Whitespace-only reasoning is not reasoning; emitting it would put
+            # an empty <think></think> in the rendered prompt, which is the
+            # exact failure this patch exists to remove.
+            if text.strip():
+                updated[_TARGET_FIELD] = text
+            out.append(updated)
+        except Exception:  # noqa: BLE001 - a request must survive a bad block
+            out.append(message)
+    return out
+
+
+__all__ = ["map_thinking_blocks_to_reasoning_content"]
--- a/config/litellm/openrouter_reasoning_roundtrip.py
+++ b/config/litellm/openrouter_reasoning_roundtrip.py
@@ -27,8 +27,37 @@ actually reads. OpenRouter accepts ``reasoning``, ``reasoning_content`` and
 this for exactly this multi-turn tool-calling case; Poolside's template reads
 ``message.reasoning`` / ``message.reasoning_content``. We send the plain string
 form, which satisfies both: ``reasoning_details`` exists to carry encrypted or
-summarised blocks, and we have neither. Anthropic ``signature`` values are
-meaningless to OpenRouter and are not forwarded.
+summarised blocks, and we have neither.
+
+Three boundaries are deliberate and worth stating outright.
+
+**The adapter's "no thinking" sentinel is not a bad shape.**
+``ChatCompletionAssistantMessage`` is a ``TypedDict``, so the adapter's
+``thinking_blocks=(blocks if len(blocks) > 0 else None)`` leaves the key
+*present and None* on every assistant turn that reasoned about nothing — the
+majority of turns on any route, and every turn on a route that returns no
+reasoning at all. That is the dominant input to this module, not an edge case:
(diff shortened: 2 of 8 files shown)


jwbron:
## Merge order with #3699: land this one first

#3699 (the 1.94.0 bump) is branched from #3695, not from this PR, so Patch 10 is not in its tree.
The two are siblings editing the same files and whichever lands second conflicts.

**Proposed order: this PR, then #3695, then #3699.** Reasons:

- This PR is based on `main` and is mergeable now. #3699 cannot land before #3695 does, so ordering
  by readiness costs nothing and unblocks the shorter path.
- #3699 is already a renumbering and re-verification pass (it retires four patches, renumbers the
  rest, and converts the test lookups). Absorbing one more patch fits work it is already doing;
  the reverse means redoing that work here against a base that has not landed.

### The conflict is wider than the patch script

Seven files are touched by both:

```
config/litellm/Dockerfile
config/litellm/patch_litellm_cache.py
docs/development/STRUCTURE.md
docs/guides/per-agent-models.md
k8s/base/litellm-deployment.yaml
tests/config/test_litellm_runtime_modules.py
tests/config/test_patch_litellm_cache.py
```

A three-way merge of `patch_litellm_cache.py` alone produces 10 conflict hunks, including the whole
`PATCHES` list (8 entries vs 10).

### What #3699 has to redo, and what is already verified

Re-verified against the actual merged tree rather than a stock tag (see #3699 for the same check
from the other side):

- Patch 10's needle matches **exactly once** in stock 1.94.0, and **still exactly once after
  #3699's eight patches have been applied**. None of them disturbs the `cache_control` plus
  `extra_body` sequence the needle spans.
- The merged set (9 patches, 5 modules) applies cleanly against a real `litellm==1.94.0` install,
  is idempotent on a second pass, and the patched file parses.
- Functionally on 1.94.0: an assistant message carrying `thinking_blocks` comes out as
  `['reasoning_content', 'role', 'tool_calls']` with `thinking_blocks` removed, and the
  `anthropic/*` route is still declined.
- The premise still holds on 1.94.0: the Anthropic adapter still parks reasoning on
  `assistant_message["thinking_blocks"]`, and OpenRouter's request path still has no consumer for
  it.

So what is left for #3699 is bookkeeping, not investigation:

1. Renumber to `Patch 9/9` and `Module 5/5` in the patch list, the script docstring, and the
   Dockerfile's staged-file list and counts (8 patches to 9, 4 modules to 5).
2. Extend the lookup-by-descriptive-tail fix to this patch's tests. Four sites in
   `tests/config/test_patch_litellm_cache.py` still match on `startswith("Patch 10/")`, which is
   exactly the positional lookup #3699 removes elsewhere; they become
   `_patch_by_description("assistant reasoning round-trip")`.

jwbron:
## Post-merge record: what this patch does, what it cannot do alone, and one sentence above that is now too strong

Landing this as a comment because the body is history now. It also contains one claim that later
measurement weakened, corrected at the bottom.

### The control, quoted rather than summarised

Three arms, identical history, tools present in every arm, `max_tokens` 800, n=4. The **only**
difference is the assistant message in the replayed history. Pre-patch it is:

```json
{"role": "assistant", "content": null,
 "tool_calls": [{"id": "call_1", "type": "function",
                 "function": {"name": "run_shell", "arguments": "{\"command\": \"ls /etc | wc -l\"}"}}]}
```

The post-patch arm adds exactly one key, which is what this patch produces:

```json
 "reasoning_content": "The user wants a count of files in /etc. I should run `ls /etc | wc -l` rather than listing everything, since only the count was asked for. After I get the number I will report it directly without re-running anything."
```

The control arm adds that same string under `content` instead of `reasoning_content`. Same text,
same token count, different field.

| arm | prior assistant message carries | turns that reasoned | median `reasoning_tokens` |
|---|---|---|---|
| pre-patch | nothing | **0/4** | 0 |
| post-patch | `reasoning_content` | **4/4** | 105 |
| control | same text as plain `content` | **0/4** | 0 |

The control is the load-bearing cell: it is the field, not the extra context. The same result holds
when the next turn also emits a tool call (0/3 against 3/3, both arms calling the tool in every
draw). Full message sequences are in the reproduction record on #3595.

### The field is honoured on the request path

Verified separately from the cells above: `reasoning_content` primed 3/3 (398, 448, 82 tokens),
`reasoning` primed 3/3 (183, 275, 232), and neither field primed 0/2. So this patch targets a field
this route actually honours when it arrives on a request. Worth recording explicitly, because an
earlier observation that "this route's field is `message.reasoning`" was **response-path only** and
does not bear on what the request path accepts. A merged patch quietly writing to a dead field is
the kind of thing found six months later.

### It still cannot fire on this traffic without a seed

Four-turn sessions, two draws each:

| arm | opener | replay | `reasoning_tokens` per turn |
|---|---|---|---|
| A | errand, tools | no | `[0,0,0,0]` `[0,0,0,0]` |
| B | errand, tools | **yes** | `[0,0,0,0]` `[0,0,0,0]` |
| C | tool-free | **yes** | `[1601,746,309,70]` `[1448,490,368,237]` |
| D | tool-free | no | `[1198,0,0,0]` `[1720,0,0,0]` |

Arm B is this patch with nothing to replay. So the original "inert on this route" claim was **right
about the outcome and wrong about the cause**: not that the model does not reason, but that nothing
ever primes it to.

### The seed, and its limit, together

The seed is cheap: an opening turn with `tool_choice: "none"` and a reasoning-shaped prompt reasons
4/4, against **0/4** for the same errand prompt with callable tools. It is a per-request parameter,
not a change to how the tools array is assembled.

**Do not take that half without this one.** Replay-forward has since been measured at **3 of 6
per-turn continuation, with chains dropping out by turn 3**, and re-seed revival is **unproven**,
because a control chain revived unaided and so nothing attributes the recovery to the re-seed.
Budget **two or three reasoning turns**, not a sustained channel.

### Correction to the body above

The body says: *"this PR is the half that keeps the seed alive once it exists."* That is too strong.
It extends a seed by a couple of turns; it does not keep it alive. The four-turn arm C figures above
were the basis for that sentence, and their monotone decay (1601, 746, 309, 70) was already pointing
at what the 3-of-6 measurement later confirmed.

### Re-verified against current `main`, post-merge

Everything in the body was verified against pre-merge trees, so none of it transferred. Re-run
against `main` at `b4f0c5c5f`, which pins `v1.94.0` and carries 10 patches and 5 modules, with this
one landing as **Patch 7/10** rather than the 9/9 predicted:

- needle matches **exactly once** in stock 1.94.0, is consumed by the patch, and the idempotency
  marker appears exactly once afterwards
- all 10 patches and 5 modules apply, second pass is a clean no-op, patched file parses
- functionally, an assistant message carrying `thinking_blocks` goes out as
  `['reasoning_content', 'role', 'tool_calls']` on `poolside/laguna-s-2.1`, and `anthropic/*` is
  still declined with `thinking_blocks` left intact

The renumber from 10 to 7 is also the case the positional test lookups would have failed silently;
the four tests here use `_patch_by_description("assistant reasoning round-trip")` and survived it.

### Not claimed

Nothing here rests on the local-window account of the livelock, which is unsupported, nor on the
synthetic period sweep, which is retracted (its stimulus instructed the model to copy, so it
measured obedience). No synthetic reproduction of the production livelock exists.

Issue #3624, opened 2026-07-25
litellm: read OpenRouter capabilities live instead of the static model-cost map


jwbron:
## Summary

`OpenrouterConfig.get_supported_openai_params` decides whether a reasoning knob is allowed through by consulting LiteLLM's **static** `model_prices_and_context_window.json`. Every OpenRouter slug egg routes is absent from that map, so the gate fails closed and the knob is dropped before the wire, silently.

OpenRouter publishes the authoritative answer live, per model, over an **unauthenticated** endpoint. We should read that instead of a bundled file that is stale by construction.

## The static map is wrong for our entire fleet

Verified against the pinned image (`ghcr.io/berriai/litellm:v1.86.2`, `config/litellm/Dockerfile:16`):

| Slug | In `model_cost` | `supports_reasoning()` |
|---|---|---|
| `openrouter/moonshotai/kimi-k3` | absent | `False` |
| `openrouter/z-ai/glm-5.2` | absent | `False` |
| `openrouter/poolside/laguna-s-2.1` | absent | `False` |
| `openrouter/deepseek/deepseek-v4-pro` | absent | `False` |
| `openrouter/moonshotai/kimi-k2.7-code` | absent | `False` |
| `openrouter/deepseek/deepseek-r1` (control) | present | `True` |

Absent is the normal state: OpenRouter adds models faster than the map tracks them.

## What the live API gives us

`GET https://openrouter.ai/api/v1/models` returns **HTTP 200 without an API key**, 345 models, each carrying `supported_parameters`:

| Model | reasoning knobs advertised |
|---|---|
| `moonshotai/kimi-k3` | `reasoning_effort`, `reasoning`, `include_reasoning` |
| `z-ai/glm-5.2` | `reasoning_effort`, `reasoning`, `include_reasoning` |
| `poolside/laguna-s-2.1` | `reasoning`, `include_reasoning` (**no** `reasoning_effort`) |

This is strictly better than the `model_info: {supports_reasoning: true}` override that #3620 documents. That override is a blunt operator assertion applied per entry; laguna shows why that matters, since asserting it there would push a parameter the model does not accept. The live lookup gets the per-model distinction right with no operator judgement involved.

A finer-grained variant exists where we need it: `GET /api/v1/models/{author}/{slug}/endpoints` returns `supported_parameters` per provider, which is the correct granularity given we pin `provider.order` on every row.

[...]

Issue #2799, opened 2026-05-27
Pilot Qwen 3.7 Max across all SDLC agents via OpenRouter + LiteLLM


jwbron:
## Motivation

Per-agent model routing via LiteLLM landed in #2769 but has not been exercised end-to-end with a non-Claude model. This issue is the first real pilot: drive a full SDLC pipeline run on **Qwen 3.7 Max** (Alibaba, May 2026) routed through **OpenRouter → our existing LiteLLM proxy**, and measure cost + quality against the current all-Opus-4.7-xhigh baseline.

## Decision: Qwen 3.7 Max for this pilot

Among models that are both materially cheaper than Opus 4.7 and competitive on coding/agentic benchmarks, **Qwen 3.7 Max is the right starting point for this pilot.** Competitors considered:

| Model | Input / Output \$/M | Key tradeoff vs Qwen |
|---|---|---|
| **Qwen 3.7 Max** | 2.50 / 7.50 | **Cheapest output of the top tier** + 1M context + strong agentic benchmarks (beats Opus 4.6 on Terminal-Bench 2.0, SWE-Bench Pro). |
| Gemini 3.1 Pro | 2 / 12 (<200K) | Slightly higher AAI score (57.2 vs 56.6) but 60% more expensive on output — bad fit for egg's output-heavy loops. |
| GPT-5.4 Standard | 2.50 / 15 | Same input as Qwen, 2× output cost. |
| DeepSeek V4 Pro Max | ~2.17 blended | Heavy reasoning tokens make effective cost similar to Qwen; doubles harness-validation surface. |
| Composer 2.5 | 0.50 / 2.50 | Cheaper still, but Cursor's model — access outside Cursor is unclear. |
| GLM-5.1 | open-weight | Strong but needs a host (self or provider) — separate workstream. |

Qwen wins on **output cost** (the dominant cost for egg's workloads), **context window** (1M, matches Opus 4.7[1m]), and **agentic-coding benchmarks**, while staying within ~1 AAI point of the next-best cheaper option. Gemini and GPT-5.4 are worth a follow-up bake-off if Qwen exposes a quality gap.

## Scope (start small)

- **Model:** Qwen 3.7 Max for **every** SDLC agent (refiner, planner, coder, tester, documenter, BRC consensus voters, overseer tiers).
- **Proxy:** the LiteLLM deployment from #2769; add an OpenRouter backend and a `qwen3.7-max` alias.
[...]

PR #3699, opened 2026-07-29
Bump litellm 1.86.2 -> 1.94.0, retire four absorbed patches (#3697)

### LINKED ISSUE #3697: Retire the litellm patch set by bumping off 1.86.2 — upstream has absorbed seven of eleven
**p2** — found while fixing #3691 (PR #3695), which added patches 10 and 11 and prompted an audit of the whole set.

`config/litellm/patch_litellm_cache.py` now carries **eleven** build-time patches against a pinned `ghcr.io/berriai/litellm:v1.86.2`. Most of them no longer need to exist: upstream has absorbed them. The image pin is what keeps us maintaining needles against code upstream has already fixed.

## Audit against upstream / the fork

Checked against `upstream/litellm_internal_staging` (v1.95.0-dev, 2026-07-28) and `jwbron/litellm@86a53b66b3`.

| egg patch | Status |
|---|---|
| 1 `CacheControlSupportedModels` | **Partial, and a disagreement.** Upstream/fork add `QWEN`; the fork *deliberately omits* `DEEPSEEK`: "its caching is automatic/prefix-based and ignores cache_control, so listing it here would be inert." If that is right, egg's DEEPSEEK entry is inert and should go. |
| 2 cache_control gate | **Absorbed, refactored** into a named `supports_cache_control_passthrough(model)` — whose docstring makes egg's exact argument about not widening `is_anthropic_claude_model` because it also gates thinking. |
| 3 `x-anthropic-billing-header` filter | **Still needed.** Upstream has it in `anthropic/chat/transformation.py` and `messages/transformation.py` but NOT in the `experimental_pass_through/adapters` path egg uses. |
| 4 `reasoning_content` thinking block | Absorbed |
| 5a/5b first-delta requeue | Absorbed |
| 6 streaming cache_read fallback | Absorbed natively (as the patch's own docstring predicted) |
| 7 live capabilities | Fork-only (jwbron/litellm#8) — **not** in BerriAI |
| 8 drop_params visibility | Fork-only (jwbron/litellm#7) — **not** in BerriAI |
| 9 thinking→reasoning_effort synthesis gate | **Nowhere.** Upstream still synthesizes for non-Claude. egg-only, and load-bearing (#3624: the derived bucket is a cap *below* the model default) |
| 10 streamed `cost` | **Upstream v1.94.0** (`8a49423`, port of BerriAI#16162) — and it goes further, feeding the real cost into litellm's own cost calculator via `_hidden_params` |
| 10 `cost_details` | Nowhere. BYOK path |
| 11 live pricing | Nowhere in BerriAI; now proposed for the fork in jwbron/litellm#10 |

## What a bump buys

Bumping to a **stock published release ≥ 1.94** retires patches 2, 4, 5a, 5b, 6, and 10-`cost` outright, plus most of 1. Seven of twelve, with no fork dependency and no change to how the image is built (`FROM` a published tag, as today).

What would remain: 3, 7, 8, 9, 10-`cost_details`, 11. Note 7, 8 and 11 are already installed *modules* rather than source needles, which is the cheap-to-maintain form — needle drift is the maintenance cost, and those have none.

## Why not just base on the fork instead

Considered and rejected for now:
- It tracks `litellm_internal_staging` (pre-release), and it is a personal fork. Different risk profile for cluster infrastructure than a pinned published release.
- It would mean building the image from source rather than `FROM` a published tag.
- It is 342 upstream commits behind, so it needs its own sync before it could be a base.

The fork remains the right home for 7/8/11 as *upstreamable* work; it is not the right base for egg's image.

## The actual cost: re-validation, not code

1.86.2 → ≥1.94 is eight minor versions of change in exactly the Anthropic-adapter surface egg depends on. Every baseline we have was established on 1.86.2:

- [ ] **Prompt-cache hit rate.** Currently ~99.99% on Qwen via OpenRouter with identical-prefix turns (#2799). Re-measure; this is the one that pays for the image existing at all.
- [ ] **Reasoning depth.** #3624 measured kimi-k3 at 3130 reasoning tokens with no parameter vs 340 with `reasoning_effort: high`, distributions non-overlapping. Patch 9 depends on that finding; confirm the adapter still derives the bucket the same way and that the gate still lands.
- [ ] **`request_params` capture.** #3599 verified `optional_params` shape on the `anthropic_messages` route, streaming and non-streaming. Re-verify the allowlist still resolves.
- [ ] **Streamed cost.** Confirm upstream's `cost` carry reaches `cost_callback._extract_cost`, and decide whether patch 10 shrinks to the `cost_details` half or is dropped.
- [ ] **Cache-token usage merge (patch 6's territory).** Confirm `cache_read_input_tokens` still reaches Claude Code, since context tracking and auto-compaction ride on it.
- [ ] Re-run the patch script against the new tree; every retired patch should fail its needle check, which is the signal to delete it rather than re-point it.

Patch 10's needle will fail the build on any bump to ≥1.94. That is by design and is the right prompt to open this work, not a bug to route around.

## Suggested sequencing

1. Land jwbron/litellm#10 (pricing + drift repair in the fork).
2. Get 7/8/11 into BerriAI so they stop being fork-local; then a bump retires them too.
3. Bump egg, delete the retired patches, re-run the checklist above.

Step 3 is the one that needs a measurement window, not just review.

jwbron:
Closes #3697. **Stacked on #3695** — it needs the two patches that PR adds. Review that one first; this targets its branch.

Bumps `ghcr.io/berriai/litellm` from **v1.86.2 to v1.94.0** (the newest stable published tag) and takes the patch set from twelve entries to eight.

## Retired — four, each verified in the 1.94.0 source

- the streaming `reasoning_content` → thinking-block branch
- the first-delta requeue on block transitions, **both** sync and async entries — upstream generalized it behind `_delta_has_content`, so it now covers text and thinking rather than only `input_json_delta`
- the `prompt_tokens_details.cached_tokens` fallback

## Narrowed — one

1.94.0 carries streamed `cost` natively (upstream `8a49423`, a port of BerriAI/litellm#16162) and goes further than egg's patch did, feeding the figure to litellm's own cost calculator. It still drops `cost_details`, which is where the BYOK number lives — OpenRouter's top-level `cost` is `0` when billing routes past them.

So patch 7 is now `cost_details` only. That needed **no code change**: the helper never overwrote a value litellm had already set, so it no-ops on `cost` and supplies the missing half. The scope narrowing is entirely documentation.

## Repointed — two, and this is the part to read

Patch 2 (the cache_control gate) and patch 5 (the drop_params warning) both stopped matching at 1.94.0 — purely because upstream **reflowed the surrounding code**. The gate collapsed onto one line and gained `is_bedrock_arn_model`; the drop site lost a line break. Neither fix was absorbed.

That matters because #3697 originally told the reader to treat a needle miss as the signal to delete. It isn't. A miss means the code moved **or** the fix landed, and the miss alone cannot tell you which.

Verified against the **stock** 1.94.0 wheel:

```
=== STOCK 1.94.0 ===
openrouter/qwen/qwen3-max         cache_control kept: 0   billing-header leaked: True
openrouter/deepseek/deepseek-v3.2 cache_control kept: 0   billing-header leaked: True
claude-sonnet-4                   cache_control kept: 1   billing-header leaked: True
openrouter/openai/gpt-4o-mini     cache_control kept: 0   billing-header leaked: True

=== PATCHED 1.94.0 ===
openrouter/qwen/qwen3-max         cache_control kept: 1   billing-header leaked: False
openrouter/deepseek/deepseek-v3.2 cache_control kept: 1   billing-header leaked: False
claude-sonnet-4                   cache_control kept: 1   billing-header leaked: False
openrouter/openai/gpt-4o-mini     cache_control kept: 0   billing-header leaked: False
```

Deleting patch 2 on the needle miss would have shipped an image paying **full input rate on every Qwen turn** — the ~10x regression the image exists to prevent — with a green build and no log line. Note the last row: gpt-4o-mini still correctly does *not* get `cache_control`. The gate is broadened, not widened.

The script docstring and the Dockerfile now both say a needle miss means "look at this", not "upstream fixed it".

I've posted a correction on #3697: my original audit was built by diffing against the **fork**, which carries its own local patches, so three fork-local fixes read as "upstream absorbed it". Redone against the stock wheel.

## Fixes a live false pass in the patch tests

The tests looked patches up by number prefix (`"Patch 4/"`). Numbers are positional, so retiring patch 4 silently re-bound that test to the openrouter capability patch — where it applied an unrelated needle to a fixture built from that same needle, and kept reading green while testing nothing.

Lookups are now by the descriptive tail of the label, which is stable across renumbering, asserting exactly one match so a retired patch fails loudly instead. The test for the retired patch 4 is deleted rather than repointed.

## Verification

All against the real 1.94.0 wheel with the patch script applied:

- all 8 patches apply cleanly and are idempotent on a second pass
- `cost` **and** `cost_details` survive stream reassembly
- `laguna-s-2.1` prices from the live rate card; `cost_per_token` matches hand arithmetic on a 100k-prompt / 90k-cached turn
- `cache_control` survives for qwen/deepseek, not for gpt-4o-mini; the billing header is filtered on every route (table above)
- the `drop_params` warning fires with the full remedy text
- the synthesized `reasoning_effort` stays suppressed by default

`make lint` clean; `tests/config/` 328 passed.

## What is NOT done

The two live measurements #3697 lists need a cluster and a model budget, and I could not run them:

- [ ] cache hit rate on real traffic (currently ~99.99% on 1.86.2, #2799)
- [ ] the #3624 reasoning-depth numbers on 1.94.0

The static behaviour both rest on is verified above — `cache_control` reaches the wire, the synthesis stays gated — but the numbers themselves are unconfirmed on the new base. **Worth a smoke run before this rolls to anything that matters.**


jwbron:
## Merge order with #3698: land that one first, then this

#3698 (the assistant reasoning round-trip, Patch 10) is branched from `main`, and this PR is
branched from #3695, so **Patch 10 is not in this tree**. The two are siblings editing the same
files, and this PR's docstring, Dockerfile and eight-patch count do not account for Patch 10.

**Proposed order: #3698, then #3695, then this PR.** #3698 is mergeable against `main` today,
whereas this PR is already gated on #3695; and this PR is already a renumbering and re-verification
pass, so absorbing one more patch fits work it is doing anyway.

### Scope of the conflict

Seven files are touched by both PRs:

```
config/litellm/Dockerfile
config/litellm/patch_litellm_cache.py
docs/development/STRUCTURE.md
docs/guides/per-agent-models.md
k8s/base/litellm-deployment.yaml
tests/config/test_litellm_runtime_modules.py
tests/config/test_patch_litellm_cache.py
```

A three-way merge of `patch_litellm_cache.py` alone gives 10 conflict hunks, the largest being the
whole `PATCHES` list (this PR's 8 entries against #3698's 10).

### Patch 10 re-verified against the merged tree, not a stock tag

Because a needle miss cannot distinguish "code moved" from "fix landed" (this PR's own point), the
check was run against the merged base rather than inherited from a stock-tag result:

| check | result |
|---|---|
| Patch 10 needle in stock 1.94.0 | 1 match |
| Patch 10 needle after this PR's 8 patches apply | **1 match** |
| merged set (9 patches, 5 modules) on a real `litellm==1.94.0` install | applies cleanly |
| second pass | idempotent, file unchanged, marker count 1 |
| patched file parses | yes |
| `transform_request` with `thinking_blocks` on 1.94.0 | emits `reasoning_content`, drops `thinking_blocks` |
| `anthropic/*` route | still declined |

The needle survives because it spans the `cache_control` pair plus the following `extra_body` pop,
and 1.94.0's reflow (the `super().transform_request(...)` collapse) falls one line below the
needle's end. None of this PR's eight patches touches that region either.

The premise also still holds on 1.94.0: the Anthropic adapter still parks prior-turn reasoning on
`assistant_message["thinking_blocks"]`, and OpenRouter's request path still has no consumer for that
field (its only reasoning lines are the `reasoning_effort` param handling and the response-path
`reasoning` to `reasoning_content` map).

### What this PR needs to add when it rebases

1. Renumber Patch 10 to `Patch 9/9` and its module to `Module 5/5`, in the patch list, the script
   docstring, and the Dockerfile (staged-file list, 8 patches to 9, 4 modules to 5).
2. Extend the lookup-by-descriptive-tail fix to Patch 10's tests. Four sites in
   `tests/config/test_patch_litellm_cache.py` still use `startswith("Patch 10/")`, the same
   positional lookup this PR removes elsewhere; they should become
   `_patch_by_description("assistant reasoning round-trip")`. Worth doing in the same pass, since a
   renumber is exactly the event that makes the positional form pass while testing nothing.

jwbron:
## Post-merge: the needle verification re-run against `main`, and what the renumber demonstrated

The merge-order half of my earlier comment is spent: this merged at 18:37Z with #3698 and #3695
ahead of it and #3700 alongside. What is still worth having on the record is whether the needle
verification survived, since every check in that comment was against a tree that no longer exists.

**It does, but the specifics moved, and the way they moved is the point.**

I predicted the reasoning round-trip patch would land as `Patch 9/9`. It landed as **`Patch 7/10`**,
because #3700 added a patch and the merge ordered them differently. Both halves of my prediction were
wrong, which is a small live demonstration of this PR's own argument: patch numbers are positional
and nothing about them is stable across a merge.

The lookup fix in this PR is what absorbed that. The four round-trip tests use
`_patch_by_description("assistant reasoning round-trip")`, so the 10 to 7 renumber moved them without
comment. Under the old `startswith("Patch 10/")` form they would have rebound to the billable-cost
gate, which now occupies `Patch 10/10`, and gone on passing while testing an unrelated needle. That
is the exact failure this PR described, and the renumber that would have triggered it happened within
hours of the fix landing.

### Re-verified against `main` at `b4f0c5c5f`

Fresh stock `litellm==1.94.0`, `main`'s own patch script, nothing inherited from the pre-merge runs:

| check | result |
|---|---|
| round-trip needle in stock 1.94.0 | 1 match |
| after the full set applies | consumed, marker present exactly once |
| all 10 patches and 5 modules | apply cleanly |
| second pass | no-op, file unchanged |
| patched file | parses |
| `transform_request` on `poolside/laguna-s-2.1` | `thinking_blocks` becomes `reasoning_content` |
| `transform_request` on `anthropic/*` | still declined, `thinking_blocks` left intact |

So the needle survived both the version bump this PR performed and the three-way merge around it,
and it survived for the reason given at the time: it spans the `cache_control` pair plus the
following `extra_body` pop, and 1.94.0's reflow of the `super().transform_request(...)` call falls
one line below the needle's end.

### One caveat on my earlier comment

It asserted the merged set was verified "against the actual merged tree". That was true of the tree I
could construct at the time, by three-way merging this branch with #3698. It was not the tree that
actually landed, which also contains #3700. The table above is against what landed.

